PR #21, opened 2020-03-05
initial Shelly2 support

eurich:
Shelly 2 in relay mode added.

Tested: 
-  manual switches both work as expected
- relays both work as expected
- Reverting back to stock firmware works as expeted..

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: all Shelly1 Shelly25
+.PHONY: all Shelly1 Shelly1PM Shelly25 Shelly2 Shelly-Plug-S
 
 MOS ?= mos
 LOCAL ?= 0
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -22,6 +22,9 @@ Shelly1PM: build-Shelly1PM
 ShellyPlugS: build-ShellyPlugS
 	@true
 
+Shelly25: build-Shelly2
```
kzyapkov: duplicate target name here, you probably meant `Shelly2`

mos.yml
```diff
@@ -1,6 +1,6 @@
 author: 'Deomid "rojer" Ryabkov'
 description: A HomeKit firmware for Shelly switches
-version: 1.3.3
+version: 1.3.4
```
rojer: this is new functionality, so make it 1.4.0 (https://semver.org/)

rojer:
thank you for contibuting! please address Kiril's comment and updater version and we're good to go.

eurich:
should be fixed now :)

rojer:
thank you!

rojer:
pushed firmware zip file

PR #60, opened 2020-04-21
Auto off delay support for down to 10ms.

### LINKED ISSUE #53: “Auto off” functionality in mili-seconds....
Hello Folks,

The 1.8 release is awesome! Thanks a lot ;o)
I face a specific application that would require pulses in mili-seconds.....
The regular Shelly firmware can be set up this way. Any chance that the next release could consider a format: like: 00:00:00:01 or 00:00:00:05 ?

Thanks for your support in advance,
Flying150!

handy-manny-swe:
Fixes #53 .

Marked as draft as it needs some more testing.

### DIFF
--- a/fs/index.html
+++ b/fs/index.html
@@ -94,7 +94,7 @@
           </div>
           <div>
             <label for="sw1_auto_off_delay">Auto off delay:</label>
-            <input type="text" id="sw1_auto_off_delay" placeholder="HH:mm:ss" required pattern="(0*([0-9]|1[0-9]|2[0-3])):(0*([0-9]|[1-4][0-9]|5[0-9])):(0*([0-9]|[1-4][0-9]|5[0-9]))">
+            <input type="text" id="sw1_auto_off_delay" placeholder="HH:mm:ss.ff" required pattern="(0*([0-9]|1[0-9]|2[0-3])):(0*([0-9]|[1-4][0-9]|5[0-9])):(0*([0-9]|[1-4][0-9]|5[0-9])).([0-9][0-9])">
           </div>
           <div class="form-control">
             <label></label>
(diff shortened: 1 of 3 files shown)


rojer:
I suggest the following pattern: `HH:MM:SS.sss`, `sss` being the millisecond part.
we won't be able to support low fractional parts but it will be more familiar, having just 2 digits is confusing.

handy-manny-swe:
> I suggest the following pattern: `HH:MM:SS.sss`, `sss` being the millisecond part.
> we won't be able to support low fractional parts but it will be more familiar, having just 2 digits is confusing.

I thought about that as well, but I personally find it more confusing to invite the user to type values that we don't support.

rojer:
there's that, but it's just not intuitive if you have just 2 digits. if it's 3, you realize it's milliseconds immediately. lower digit won't be meaningful, but for presentation sake let's have 3.

handy-manny-swe:
> there's that, but it's just not intuitive if you have just 2 digits. if it's 3, you realize it's milliseconds immediately. lower digit won't be meaningful, but for presentation sake let's have 3.

Well, it's only milliseconds in code. That does not necessarily need to be reflected in the UI as we could as well call it hundredths of second. But we'll go with 3 and make some kind of effort to not trick the user into typing values less than 10ms.

andyblac:
@handy-manny-swe or just code it so anything less than 10ms is actioned as 10ms

handy-manny-swe:
> @handy-manny-swe or just code it so anything less than 10ms is actioned as 10ms
> 
> and display a note on screen say so.

I'm leaning towards validation and validation message. At least that's what I've just coded. :)

andyblac:
i wish i knew C so i could help you guys 😄

handy-manny-swe:
Seems to work. My one and only shelly1 is installed in my garage door electronics box, so my testing possibilities are a bit limited. But a setting of 10ms sure triggers a lot faster than a setting of 1s.

rojer:
merged, with some minor modifications

PR #63, opened 2020-04-25
Bug fixes for validation pattern for auto-off delay

handy-manny-swe:
Fixes some issues with the validation pattern for auto-off delay.
Before this fix these issues exists
- possible to provide auto off delay of zero
- possible to enter values such as 023:059:059.000, which doesn't make sense.

This fix ensures that at least one of the following is true

- hours is between 1 and 23
- minutes is between 1 and 59
- seconds is between 1 and 59
- milliseconds is between 10 and 999


rojer:
this makes my eyes bleed even more :)
https://xkcd.com/208/ comes to mind

ok, even with what we had previously, we were way past the point where regular expressions lose their usefulness.
so let's leave a basic regex that enforces that each group consists of numbers and do the rest in JS.

andyblac:
and besides, one could need more than 24 hours, ie timed off if you go on hols, turn off xxx after xx days.

handy-manny-swe:
In that case we should probably go for minimum 10ms and maximum... uhmm whatever the mongoose os timer max value is and keep it in js as @rojer suggested.

andyblac:
until a schedule is coded, ie turn off on date / time, then a max of 23:59:59.999 would be fine imo

handy-manny-swe:
> until a schedule is coded, ie turn off on date / time, then a max of 23:59:59.999 would be fine imo

That would mean something like ((0[0-9]|1[0-9]|2[0-3])):([0-5][0-9]):([0-5][0-9]).([0-9][0-9][0-9]) as far as regex goes.

handy-manny-swe:
> this makes my eyes bleed even more :)
> https://xkcd.com/208/ comes to mind
> 
> ok, even with what we had previously, we were way past the point where regular expressions lose their usefulness.
> so let's leave a basic regex that enforces that each group consists of numbers and do the rest in JS.

I prefer this one :)
https://xkcd.com/1171/

rojer:
let's keep it simple, regex-wise - `[0-9]+:[0-9]{2}:[0-9]{2}:[0-9]{2}\.[0-9]{3}` or something to that tune, - and to the rest in JS.

handy-manny-swe:
Given comments from @rojer and @andyblac the scope changed to support days as well. This is now implemented. 

The input validation is divided into two parts: pattern/regex for pattern matching, JS for range validation.

rojer:
merged, thank you!

PR #122, opened 2020-08-31
[FlashScript] check for local domain, add automatically if required

andyblac:
OK, added like you requested.


rojer:
thanks!

andyblac:
np, anything else just let me know.

PR #134, opened 2020-09-14
[FlashScript] add -V so you can force a firmware version.

andyblac:
usage:
./flash_shelly.sh -f 2.1.2 hostname

#To Do
add support to force flash Stock Firmware Versions.

### DIFF
--- a/tools/flash_shelly.sh
+++ b/tools/flash_shelly.sh
@@ -231,8 +231,13 @@ function probe_info {
           *) ;;
         esac
       fi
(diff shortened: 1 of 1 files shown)


rojer:
`-f` usually stands for "force", as in "bypass safety checks, do what i say".
`-v` is usually "verbose output", let's use "-V".

andyblac:
> `-f` usually stands for "force", as in "bypass safety checks, do what i say".
> `-v` is usually "verbose output", let's use "-V".

ok, what flag would you recommend ?

andyblac:
oops ok `-V`

PR #135, opened 2020-09-14
[FlashScript] refactor "Waiting for device to reboot"

andyblac:
This removes a simple sleep to a proper wait loop with a max of 30 seconds, before reporting a failure.


rojer:
thanks!

PR #137, opened 2020-09-19
Version Check and GUI Update

schemhad:
* Adds a VersionCheck for major Releases over GitHub API
* Adds Favicon
* Dynamically sets Hostname as PageTitle
* Slightly GUI improvements


rojer:
can you add screenshots / screencap of the process?

schemhad:
image

schemhad:
image

The function compares the latest tag_Name only from the releases-api with the local one.

andyblac:
@schemhad could you the ability to update like we discussed in the ticket ?

schemhad:
@andyblac I think so, but only with major releases that are in the release section / api listed. 

And then we have to determine the device type (1/2/PM/PlugS/...) so select the corresponding URL, fetch the hole update file and paste it to the /update function. 

Should be possible, but takes a little time ...

rojer (changes requested):

andyblac:
> @andyblac I think so, but only with major releases that are listed in the release section / api.
> 
> And then we have to determine the device type (1/2/PM/PlugS/...) so select the corresponding browser_download_url, fetch the hole update file and paste it to the /update function.
> 
> And we could switch between major-releases and testing-releases.
> 
> The function in this pull request is for "normal" users, so they could see if there is a major(stable)-release.
> 
> Should all be possible, but takes a little time ...

ye, major releases should be fine, we have my script for flashing test versions. regarding device type thats in `http://xxx.xxx.xxx.xxx/rpc/Shelly.GetInfo` under `app`
`{"id": "shelly1-F2824A", "app": "switch1", "model": "Shelly1", "host": "shelly1-F2824A.local",` that is tagged in GitHub downloads

schemhad:
New PullRequest incoming, this time with more time, love and testing.

Issue #65, opened 2020-04-28
hostname change


andyblac:
is it possible to edit the hostname, like you did you switch names ?

![Screenshot 2020-04-28 at 19 45 14](https://user-images.githubusercontent.com/77936/80525595-679ba580-8989-11ea-97a7-20f94af1a441.jpg)

rojer:
yes, it is possible. it is specified in `device.id` config variable

MilanBacaCz:
and will it be accessible from config gui?

andyblac:
@rojer could this be added, be handy when you have multiple shelly's for the script to quickly see which is which.

andyblac:
@rojer does this look correct ?, if so I'll create a pull request 

https://github.com/andyblac/shelly-homekit/commit/eb66793be77a97ac83190a3c8278659fd6318b66

how can I test such things locally ?

andyblac:
@schemhad if you have time could take a look at this ?

schemhad:
That's why i've set the pagetitle dynamically in the recent pull request. 😉

But i would place it in the wifi container.

andyblac:
does my change look correct (just move to wifi), I'm new as JS, just trying to understand from existing code.

rojer:
> @rojer does this look correct ?, if so I'll create a pull request
> 
> [andyblac@eb66793](https://github.com/andyblac/shelly-homekit/commit/eb66793be77a97ac83190a3c8278659fd6318b66)

not quite. this needs to be sent to the device as Device.SetConfig RPC, and the device needs to apply it to the config.
a small code change is needed.

> 
> how can I test such things locally ?

there's a way to iterate on web changes without reflashing the device.
1) make the changes to `fs_src/index.html`
2) compress it:

```
[rojer@nbd ~/allterco/shelly-homekit master]$ make fs/index.html.gz
gzip -9 -c fs_src/index.html > fs/index.html.gz
```

3) upload to the device using the `mos` tool:

```
$ mos --port ws://shelly1-68b743.local/rpc put fs/index.html.gz
```

4) refresh the webpage in the browser, test the changes

rojer:
https://github.com/mongoose-os-apps/shelly-homekit/commit/95c21e455b836f81bdf320ce28ddbec400484d9e adds the API feature, @schemhad please add GUI when you have the time.

rojer:
i was adding system settings section for roller-shutter and added hostname item to it.

[...]

PR #148, opened 2020-09-24
Version Check & GUI

schemhad:
* Adds a simple VersionCheck for major Releases over GitHub API
* Adds notification Badge on firmware section header
* Adds Favicon
* Dynamically sets Hostname as PageTitle
* Slightly GUI improvements


schemhad:
image

image

andyblac:
Dose add support for changing host name ?

schemhad:
Not for now, there are too many changes already. 

If this one is merged, the hostname is the next thing on my list.

rojer:
it's best to keep pushing to the same pull request branch rather than creating new one each time.
so, as i said on the previous request - there's only a short step to actually flashing the firmware, can you try doing that? it would be ideal.
otherwise, can you at least add a link to the release so user can go directly there?

rojer:
> Not for now, there are too many changes already.
> 
> If this one is merged, the hostname is the next thing on my list.

that's great. a small change on the device side.
@andyblac can you please file a separate issue for that so we can coordinate?

andyblac:
> > Not for now, there are too many changes already.
> > If this one is merged, the hostname is the next thing on my list.
> 
> that's great. a small change on the device side.
> @andyblac can you please file a separate issue for that so we can coordinate?

Iirc there is one already

rojer:
ah, true, #65

schemhad:
> it's best to keep pushing to the same pull request branch rather than creating new one each time.
> 
> so, as i said on the previous request - there's only a short step to actually flashing the firmware, can you try doing that? it would be ideal.
> 
> otherwise, can you at least add a link to the release so user can go directly there?

I'm afraid that I'll lose Track on the changes I already Made. After every commit i have to start with copy, paste and test. If we could merge this one, not to a new release, but just to the latest commit, it will help me. 

After that, I'll bring up the direct download link ASAP and take a look at fetching the whole update and change the hostname.

rojer:
> After every commit i have to start with copy, paste and test. If we could merge this one, not to a new release, but just to the latest commit, it will help me.

that sounds painful. are you using Git on your machine at all at all? if not, you should.

i sympathize with your troubles, but merging half-baked changes is not the way to go.
master branch is supposed to be release-ready at any time. development is done on different branches (you created one by creating a pull request, for example), and integrated into master when they are release-ready.

read up on Git, i'm sure GitHub has docs. it's good for you and will make your life easier.
there's a version for windows with GUI as well, if you like that sort of thing - https://git-scm.com/downloads

andyblac:
ye, all I do I clone master, take a branch locally to work on, pushing to local git, then when I'm read if no changes on master have been done, I just push it to remote and create a  Pull Request, if master has had new commits, I use pull those in, create another branch of it, and cherry pick in my changes, to that new branch to remote and create a PR.

andyblac:
try sourcetree thats what I use. https://www.sourcetreeapp.com. just right click on a branch at left and choose create Pull Request it really simple to use.

Screenshot 2020-09-25 at 10 16 35

schemhad:
image

Now with download device-specific ZIP-File.

schemhad:
"**Download for %deviceType% >**" will resolve to the latest version for corresponding Device ("Browser_Download_URL" over GitHub API).

Please ignore the wrong verison numbers in the screenshot.

schemhad:
Please give it a try.

andyblac:
branch has conflicts can't be merged.

schemhad:
Ok, sorry. I didn't see that. Now it could be merged.

schemhad:
It happens everytime i zip the sources and update the index.html.gz.

rojer:
small request: since you're pulling release information anyway, can you display release notes in a text area below?

i'll merge it and try to add post'ing myself so we don't need to have it saved on the computer at all.

schemhad:
I've got an Alert with Changelog of the last 'n' version's nearly done. Is an default javascript Alert ok? I think it would be better for the overview...

But could we merge the actual state first?

rojer:
ok, let's get it in.

>  Is an default javascript Alert ok? I think it would be better for the overview...

i'd rather it be a <div> that opens below the download link, alerts are very intrusive and are only (barely) acceptable for error indication.

andyblac:
just looking at beta4, could you add a checkbox to allow beta updates to be checked for ?, so user can decide if they want to use them or not ?

rojer:
maybe in the future.
for now, i think it's fine to load first beta manually - from a link on the issue, presumably.
i'd like users to have context around what's being tested first.

andyblac:
> maybe in the future.
> for now, i think it's fine to load first beta manually - from a link on the issue, presumably.
> i'd like users to have context around what's being tested first.

ah, that fine then, if you have to load beta manually, then after that it will check for new beta's, I'll look into doing the same for my script.

rojer:
so, the way it works in the web UI since 2.2 is:
 1. we fetch https://rojer.me/files/shelly/update.json
 2. go through the list applying regexes (first item of the tuple) to our version string
 3. when one that matches is found, we check against `version` field of the section (second tuple) and if it's newer, we consider it an update.

ideally, script should implement the same algorithm.

andyblac:
so drop GitHub tags ?

rojer:
yeah, pretty much. i added this json config for more flexibility - you can see there now the section i was using for beta updates: it's match 2.2.0beta-* and give them the next beta, while giving latest (2.1.8 at the time) to everyone else. then i disabled it, so everyone, including former beta users, now get 2.2.0 release.

andyblac:
having difficulty getting tag name ie

```
        lfw=$(echo "$homekit_release_info" | jq -r .tag_name)
        dlurl=$(echo "$homekit_release_info" | jq -r '.assets[] | select(.name=="shelly-homekit-'$model'.zip").browser_download_url')
```

```
[
  [
    "X^2.2.0-beta.*", {
      "version": "2.2.0-beta8",
```

how do I select the 1st tag "X^2.2.0-beta.*"

rojer:
honestly, i think we're pushing beyond what shell is capable of here.
how do you feel about rewriting the script in Python? mac os has it installed by default, so does linux. and windows doesn't have bash either, so who cares :)

rojer:
this will also obviate the need for for `jq`, `curl` or timeout - you can use Python's builtin json parser, urllib and... i guess some internal timeout magic?

andyblac:
ok yeah, I'll take a look, that would also add windows to supported os

PR #157, opened 2020-09-29
Fix typo in Shelly's device url

erezkoren:


rojer:
yes, we should remove these. Andy, please send a PR.

PR #158, opened 2020-09-29
help for non-responsive devices

erezkoren:
close it if u don't think it needed

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,9 @@ Reverting to stock firmware is also supported (see below).
 
  Script [here](tools/flash_shelly.sh) for an automated way to revert your devices to stock firmware.
    * ./flash_shelly.sh -m revert hostname
-
+ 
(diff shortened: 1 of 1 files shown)


andyblac:
We are migrating to wiki please feel free to suggest any changes on that, I’ll be happy to take a look

rojer:
yes, i've just removed almost everything from the readme in favor of wiki.
we should definitely have a section there for recovering the device - reset wifi via input toggles in the first minute, reflashin instructions, etc.

andyblac:
> yes, i've just removed almost everything from the readme in favor of wiki.
> we should definitely have a section there for recovering the device - reset wifi via input toggles in the first minute, reflashin instructions, etc.

already in WiKi, does it need any further info https://github.com/mongoose-os-apps/shelly-homekit/wiki/Recovery

PR #176, opened 2020-10-05
[FlashScript] initial upload of Python version.

andyblac:
please give through test, seems ok here.

### DIFF
--- a/tools/flash_shelly.py
+++ b/tools/flash_shelly.py
@@ -201,6 +201,8 @@ def probe_info(device, action, dry_run, silent_run, mode, exclude, exclude_devic
   host = host.replace('.local','')
 
   logger.info("\n\ndevice: %s" % device)
+  for i, item in enumerate(exclude_device):
+    exclude_device[i] = exclude_device[i].replace('.local','')
   logger.info("host: %s" % host)
   logger.info("action: %s" % action)
   logger.info("dry_run: %s" % dry_run)
--- a/tools/flash_shelly.py
+++ b/tools/flash_shelly.py
@@ -129,9 +129,9 @@ def write_flash(device, lfw, dlurl, cfw_type, mode):
   host = device.replace('.local','')
   if cfw_type == 'homekit':
     print("Downloading Firmware...")
+    logger.info('DURL: %s' % dlurl)
     myfile = requests.get(dlurl)
     open('shelly-flash.zip', 'wb').write(myfile.content)
-    # flashcmd="curl -qsS -F file=@shelly-flash.zip http://$device/update"
     if path.exists('shelly-flash.zip') or cfw_type == 'stock':
       print("Now Flashing...")
       files = {
@@ -356,12 +356,10 @@ def device_scan(args, action, do_all, dry_run, silent_run, mode, exclude, forced
 
   if do_all == False:
     print(WHITE + "Probing Shelly device for info..." + NC)
-    logger.info("probe_info: 1")
     if  not ".local" in device:
       device = device + ".local"
     probe_info(device, action, dry_run, silent_run, mode, forced_version, ffw)
   else:
-    logger.info("probe_info: 2")
     print(WHITE + "Scanning for Shelly devices..." + NC)
     zeroconf = Zeroconf()
     listener = MyListener()
--- a/tools/flash_shelly.py
+++ b/tools/flash_shelly.py
@@ -1,9 +1,17 @@
-import getopt, importlib, platform, urllib, json, logging, re
-from sys import argv, exit
-from os import path, remove
-from subprocess import Popen, PIPE
-from time import sleep
+import getopt
+import importlib
+import json
+import logging
+import os
+import platform
+import re
+import subprocess
+import sys
+import time
+import urllib
+
 from importlib import util
+from sys import argv
 
 """
 #  Copyright (c) 2020 Andrew Blackburn & Deomid "rojer" Ryabkov
--- a/tools/flash_shelly.py
+++ b/tools/flash_shelly.py
@@ -245,7 +245,7 @@ def probe_info(device, action, dry_run, silent_run, mode, exclude, exclude_devic
     if mode == 'homekit':
       model = info['model'] if 'model' in info else shelly_model(type, mode)
       for i in homekit_release_info:
-        if re.search(i[0][0], cfw):
+        if re.search(i[0], cfw):
           lfw = i[1]['version']
           if forced_version == False:
             dlurl = i[1]['urls'][model]
@@ -262,7 +262,7 @@ def probe_info(device, action, dry_run, silent_run, mode, exclude, exclude_devic
     if mode == 'homekit':
       model = shelly_model(type, mode)
       for i in homekit_release_info:
-        if re.search(i[0][0], cfw):
+        if re.search(i[0], cfw):
           lfw = i[1]['version']
           if forced_version == False:
             try:
--- a/tools/flash_shelly.py
+++ b/tools/flash_shelly.py
@@ -39,9 +39,9 @@ import functools
 import getopt
 import importlib
 import importlib.util
+import io
 import json
 import logging
-import os
 import platform
 import re
 import subprocess
@@ -168,12 +168,12 @@ def write_flash(device, lfw, dlurl, cfw_type, mode):
     logger.debug('DURL: %s' % dlurl)
     myfile = requests.get(dlurl)
     with open('shelly-flash.zip', 'wb') as f:
-      f.write(myfile.content)
-    if os.path.exists('shelly-flash.zip') or cfw_type == 'stock':
-      logger.info("Now Flashing...")
-      files = {'file': ('shelly-flash.zip', open('shelly-flash.zip', 'rb'))}
-      response = requests.post('http://%s/update' % device , files=files)
-      logger.debug(response.text)
+      in_memory = io.BytesIO(myfile.content)
+    logger.info("Now Flashing...")
+    files = {'file': ('shelly-flash.zip', in_memory)}
+    response = requests.post('http://%s/update' % device , files=files)
+    logger.debug(response.text)
+
   else:
     logger.info("Now Flashing...")
     dlurl = dlurl.replace('https', 'http')
@@ -207,17 +207,14 @@ def write_flash(device, lfw, dlurl, cfw_type, mode):
     onlinecheck = info['fw'].split('/v')[1].split('@')[0]
   if onlinecheck == lfw:
     logger.info(GREEN + "Successfully flashed %s to %s\033[0m" % (host, lfw))
-    if mode == 'homekit':
-      os.remove('shelly-flash.zip')
-    else:
-      if info['type'] == "SHRGBW2":
-        logger.info("\nTo finalise flash process you will need to switch 'Modes' in the device WebUI,")
-        logger.info(WHITE + "WARNING!!" + NC + "If you are using this device in conjunction with Homebridge it will")
-        logger.info("result in ALL scenes / automations to be os.removed within HomeKit.")
-        logger.info("Goto http://$device in your web browser")
-        logger.info("Goto settings section")
-        logger.info("Goto 'Device Type' and switch modes")
-        logger.info("Once mode has been changed, you can switch it back to your preferred mode.")
+    if mode == 'homekit' and info['type'] == "SHRGBW2":
+      logger.info("\nTo finalise flash process you will need to switch 'Modes' in the device WebUI,")
+      logger.info(WHITE + "WARNING!!" + NC + "If you are using this device in conjunction with Homebridge it will")
+      logger.info("result in ALL scenes / automations to be removed within HomeKit.")
+      logger.info("Goto http://$device in your web browser")
+      logger.info("Goto settings section")
+      logger.info("Goto 'Device Type' and switch modes")
+      logger.info("Once mode has been changed, you can switch it back to your preferred mode.")
   else:
     logger.info(RED + "Failed to flash %s to %s\033[0m" % (host, lfw))
     logger.info("Current: %s" % onlinecheck)
--- a/tools/flash_shelly.py
+++ b/tools/flash_shelly.py
@@ -141,10 +141,9 @@ def write_flash(device, lfw, dlurl, cfw_type, mode):
       logger.info(response.text)
   else:
     print("Now Flashing...")
-    params = (
-        ('url', dlurl),
-    )
-    response = requests.get('http://%s/ota' % device, params=params)
+    dlurl = dlurl.replace('https', 'http')
+    logger.info("curl -qsS http://%s/ota?url=%s" % (device, dlurl))
+    response = requests.get('http://%s/ota?url=%s' % (device, dlurl))
     logger.info(response.text)
   n = 1
   waittextshown = False
--- a/tools/flash_shelly.py
+++ b/tools/flash_shelly.py
@@ -22,7 +22,7 @@ from importlib import util
 #  limitations under the License.
 #
 #  This script will probe for any shelly device on the network and it will
-#  attemp to update them to the latest firmware version available.
+#  attempt to update them to the latest firmware version available.
 #  This script will not flash any firmware to a device that is not already on a
 #  version of this firmware, if you are looking to flash your device from stock
 #  or any other firmware please follow instructions here:
--- a/tools/flash_shelly.py
+++ b/tools/flash_shelly.py
@@ -139,9 +139,7 @@ def parseVersion(vs):
         break
       i += 1
     varSeq = int(pp[1][i]) or 0
-  major = int(v[0])
-  minor = int(v[1])
-  patch = int(v[2])
+  major, minor, patch = [int for e in v]
   return (major, minor, patch, variant, varSeq)
 
 def isNewer(v1, v2):
--- a/tools/flash_shelly.py
+++ b/tools/flash_shelly.py
@@ -190,7 +190,7 @@ def write_flash(device, lfw, dlurl, cfw_type, mode):
     print(RED + "Failed to flash %s to %s\033[0m" % (host, lfw))
     print("Current: %s" % onlinecheck)
 
-def probe_info(device, action, dry_run, silent_run, mode, forced_version, ffw):
+def probe_info(device, action, dry_run, silent_run, mode, exclude, exclude_device, forced_version, ffw):
   flash = False
   info = None           # firmware versions info
   model = None          # device model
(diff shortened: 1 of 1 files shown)


tools/flash_shelly.py
```diff
@@ -0,0 +1,499 @@
+import getopt, importlib, platform, urllib, json, logging, re
+from sys import argv, exit
+from os import path, remove
+from subprocess import Popen, PIPE
+from time import sleep
+from importlib import util
+
+"""
```
rojer, line 8: no need to make it a string. put it above imports and put `#!/usr/bin/env python3` as the first line

tools/flash_shelly.py
```diff
@@ -0,0 +1,499 @@
+import getopt, importlib, platform, urllib, json, logging, re
```
rojer, line 1: 1 import per line and sort them

tools/flash_shelly.py
```diff
@@ -0,0 +1,499 @@
...
+from sys import argv, exit
```
rojer, line 2: don't use from foo import blah, import foo and use foo.blah

tools/flash_shelly.py
```diff
@@ -0,0 +1,499 @@
...
+YELLOW='\033[1;93m' # Bright Yellow
+BLUE='\033[1;94m' # Bright Blue
+PURPLE='\033[1;95m' # Bright Purple
+NC='\033[0m' # Normal Color
+
+arch = platform.system()
+
+if not importlib.util.find_spec("zeroconf"):
+  print('Installing zeroconf...')
+  pipe = Popen('pip3 install zeroconf', shell=True, stdout=PIPE).stdout
```
rojer, line 62: `subprocess.check_output`

tools/flash_shelly.py
```diff
@@ -0,0 +1,499 @@
...
+  print('Installing requests...')
+  pipe = Popen('pip3 install requests', shell=True, stdout=PIPE).stdout
+  out = (pipe.read())
+  import requests
+else:
+  import requests
+
+if not importlib.util.find_spec("packaging"):
+  print('Installing packaging...')
+  pipe = Popen('pip3 install packaging', shell=True, stdout=PIPE).stdout
```
rojer, line 78: extra dependency, let's not use it. write python version of the js function instead.

tools/flash_shelly.py
```diff
@@ -0,0 +1,499 @@
...
+  out = (pipe.read())
+  from packaging import version
+else:
+  from packaging import version
+
+class MyListener:
+  global device_list, p_list
```
rojer, line 85: a global class variable? i didn't know you could do that. in any case, avoid global variables please, pass things around as needed or make them instance variables

tools/flash_shelly.py
```diff
@@ -0,0 +1,499 @@
...
+               'switch2' : 'SHSW-21',
+               'switch25' : 'SHSW-25',
+               'shelly-plug-s' : 'SHPLG-S',
+               'dimmer1' : 'SHDM-1',
+               'rgbw2' : 'SHRGBW2',
+    }
+  return(options[type])
+
+def write_flash(device, lfw, dlurl, cfw_type, mode):
+  logger.info("\n" + WHITE + "write_flash" + NC)
```
rojer, line 127: you use both logger and prints for diagnostic output, use one or the other. my preference is logger.

tools/flash_shelly.py
```diff
@@ -0,0 +1,499 @@
...
+  flashed = False
+  host = device.replace('.local','')
+  if cfw_type == 'homekit':
+    print("Downloading Firmware...")
+    logger.info('DURL: %s' % dlurl)
+    myfile = requests.get(dlurl)
+    open('shelly-flash.zip', 'wb').write(myfile.content)
```
rojer, line 134: this leaks an open file object.
use context managers on files, i.e.
```
with open(...) as f:
  ... do stuff ...
```

tools/flash_shelly.py
```diff
@@ -0,0 +1,499 @@
...
+
+def probe_info(device, action, dry_run, silent_run, mode, exclude, exclude_device, forced_version, ffw):
+  flash = False
+  info = None           # firmware versions info
+  model = None          # device model
+  lfw = None            # latest firmware available
+  cfw = None            # current firmware on device
+  cfw_type = 'homekit'  # current firmware type
+  host = device
+  host = host.replace('.local','')
```
rojer, line 202: always put spaces after commas in argument lists

tools/flash_shelly.py
```diff
@@ -0,0 +1,499 @@
...
+  if ((not opts and not args) and do_all == False) or (not args and do_all == False) or \
+     (args and do_all == True and exclude == False) or (not args and do_all == True and exclude == True):
+    usage()
+    exit(1)
+
+  try:
+    fp = urllib.request.urlopen("https://api.shelly.cloud/files/firmware")
+    global stock_release_info
+    stock_release_info = json.load(fp)
+    fp.close()
```
rojer, line 467: instead of closing explicitly, use context manager:

with urllib.request.urlopen(...) as fp:
 ...

tools/flash_shelly.py
```diff
@@ -0,0 +1,499 @@
...
+  if mode == 'keep':
+    mode = cfw_type
+
+  if cfw_type == 'homekit':
+    type = info['app']
+    cfw = info['version']
+    if mode == 'homekit':
+      model = info['model'] if 'model' in info else shelly_model(type, mode)
+      for i in homekit_release_info:
+        if re.search(i[0][0], cfw):
```
rojer, line 244: shouldn't this be i[0]? i think i[0][0] is the first character of the regex

tools/flash_shelly.py
```diff
@@ -0,0 +1,499 @@
...
+    fp = urllib.request.urlopen("https://api.shelly.cloud/files/firmware")
+    global stock_release_info
+    stock_release_info = json.load(fp)
+    fp.close()
+  except:
+    logger.warning("Failed to lookup version information")
+    exit(1)
+  try:
+    fp = urllib.request.urlopen("https://rojer.me/files/shelly/update.json")
+    global homekit_release_info
```
rojer, line 473: no global variables please, function arguments or instance variables.

tools/flash_shelly.py
```diff
@@ -0,0 +1,499 @@
...
+  if do_all == True and exclude == True:
+    device_scan(args, action, do_all, dry_run, silent_run, mode, exclude, forced_version, ffw)
+
+
+
+
+if __name__ == '__main__':
+  app(argv)
+
+__all__ = ['app']
```
rojer, line 499: why is this needed?

rojer (changes requested):

tools/flash_shelly.py
andyblac: that stop windows from running script.

tools/flash_shelly.py
andyblac: I normally code like above, but ok, np

tools/flash_shelly.py
andyblac: ok, guess memory is not an issue.

tools/flash_shelly.py
andyblac: no it not that, I use it here 
```
    elif (version.parse(cfw) < version.parse(lfw)) and ((cfw_type == 'homekit' and mode == 'homekit') or (cfw_type == 'stock' and mode == 'stock') or mode == "keep"):
```

tools/flash_shelly.py
andyblac: I use it in device_scan , can't pass it there 'chicken and egg'.

andyblac: but I'll try rethink code.

tools/flash_shelly.py
andyblac: you can't, that why logging states use print for normal running code.

tools/flash_shelly.py
andyblac: nope.

tools/flash_shelly.py
andyblac: runs the script from cmd line.

tools/flash_shelly.py
andyblac: I use 2 levels of debug, we only have `info` and `debug` to use, so leaves nothing for standard logging.

tools/flash_shelly.py
andyblac: this needs to be a global, as listener can not pass back data to previous def.

tools/flash_shelly.py
andyblac: done https://github.com/mongoose-os-apps/shelly-homekit/pull/176/commits/f65925f50d7ca9c6850d5735ace7621795b4e933

tools/flash_shelly.py
andyblac:
```
from importlib import util
from sys import argv
```

these 2 are required, they can not be full imports. else modules is not found.

```
Andrews-MacBook-Pro:tools andrewblackburn$ python3 flash_shelly.py -la
Traceback (most recent call last):
  File "flash_shelly.py", line 68, in <module>
    if not importlib.util.find_spec("zeroconf"):
AttributeError: module 'importlib' has no attribute 'util'

Andrews-MacBook-Pro:tools andrewblackburn$ python3 flash_shelly.py -la
  File "flash_shelly.py", line 400
    def app(sys.argv):
               ^
SyntaxError: invalid syntax
```

andyblac: https://github.com/mongoose-os-apps/shelly-homekit/pull/176/commits/a39399332926dd2f642773d4765e5c36df3396fa
https://github.com/mongoose-os-apps/shelly-homekit/pull/176/commits/5b81af09f69a18d8fde32f0891a5b4381ec425d2

tools/flash_shelly.py
andyblac: done.

tools/flash_shelly.py
andyblac: done 
https://github.com/mongoose-os-apps/shelly-homekit/pull/176/commits/5c0a6611a21db479e0022bd105492863482a5220

tools/flash_shelly.py
andyblac: like

```
    myfile = requests.get(dlurl)
    with open('shelly-flash.zip', 'wb') as f:
      write(myfile.content)
```

tools/flash_shelly.py
andyblac: done https://github.com/mongoose-os-apps/shelly-homekit/pull/176/commits/249ad1373b919b4a56bbb8355216dada8c2ce975

tools/flash_shelly.py
andyblac: done https://github.com/mongoose-os-apps/shelly-homekit/pull/176/commits/eeb8bae51752f3225a1425c7cec00adcb75675a3

tools/flash_shelly.py
andyblac: oh you where correct close.
```
re.search(i[0][0]: <re.Match object; span=(0, 1), match='2'>
```
I'll tweak to
```
re.search(i[0]: <re.Match object; span=(0, 5), match='2.2.1'>
```

andyblac: done https://github.com/mongoose-os-apps/shelly-homekit/pull/176/commits/2abe8439bc41b318b9192a0ee5c6aaacd23340c8

tools/flash_shelly.py
andyblac: done https://github.com/mongoose-os-apps/shelly-homekit/pull/176/commits/429770c217c5c98fe703eed838b2136debbd0185

tools/flash_shelly.py
andyblac: it required to pass args and opts to getopt, they are stored in the import `argv `

tools/flash_shelly.py
andyblac: done https://github.com/mongoose-os-apps/shelly-homekit/pull/176/commits/5176422fad7ccf4ab1a55b93d76c7dea76b4cdaf

tools/flash_shelly.py
andyblac: ok, reword to Instance Variables https://github.com/mongoose-os-apps/shelly-homekit/pull/176/commits/ab391f6a2cd1fa5742591296f615544503758d27

tools/flash_shelly.py
andyblac: https://docs.python.org/3/howto/logging.html

it states the following, which is what I am doing.
Screenshot 2020-10-07 at 01 03 50

andyblac: ok, I have managed to add a custom log level, tomorrow I'll replace print() with logging.

andyblac: done https://github.com/mongoose-os-apps/shelly-homekit/pull/176/commits/d444400fdda05dc1ef91777762dfa621309da183

tools/flash_shelly.py
andyblac: ie

```
if __name__ == "__main__":
   print("File one executed when ran directly")
else:
   print("File one executed when imported")
```

andyblac: I can remove `__all__ = ['app']` now we are not using `from xxx import yyyy`

andyblac: https://github.com/mongoose-os-apps/shelly-homekit/pull/176/commits/5286c45882aa179560b71b4ab5957a26d3397d00

tools/flash_shelly.py
andyblac: https://github.com/mongoose-os-apps/shelly-homekit/pull/176/commits/215e8e65fb1b07da10257e558cfad61f55ace32e

andyblac: ok after reading up, it seems it won't effect windows at all, so I'll add it, thx

tools/flash_shelly.py
andyblac: https://github.com/mongoose-os-apps/shelly-homekit/pull/176/commits/ae906910935bd41b57675857f7d42251938dad4c

tools/flash_shelly.py
andyblac: https://github.com/mongoose-os-apps/shelly-homekit/pull/176/commits/ae906910935bd41b57675857f7d42251938dad4c

tools/flash_shelly.py
andyblac: https://github.com/mongoose-os-apps/shelly-homekit/pull/176/commits/f9c90340cd36225fb2190d643fb8628d2b882543

andyblac: should be resolved now.

tools/flash_shelly.py
andyblac: https://github.com/mongoose-os-apps/shelly-homekit/pull/176/commits/f9c90340cd36225fb2190d643fb8628d2b882543

tools/flash_shelly.py
```diff
@@ -0,0 +1,501 @@
...
+logging.addLevelName(logging.TRACE, 'TRACE')
+logging.Logger.trace = functools.partialmethod(logging.Logger.log, logging.TRACE)
+logging.trace = functools.partial(logging.log, logging.TRACE)
+logging.basicConfig(format='%(message)s')
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+arch = platform.system()
+# Windows does not support acsii colours
+WHITE = '\033[1m' if not arch.lower().startswith('win') else ""
```
rojer, line 62: actually, this depends on terminal capabilities, not OS. maybe there's a way to figure it out.
but in any case, - avoid repetition where possible:

```
if not arch.lower().startswith('win'):
  ...
else:
  WHITE = RED = ... = NC = ""
```

tools/flash_shelly.py
```diff
@@ -0,0 +1,501 @@
...
+    i = 0
+    for x in pp[1]:
+      c = pp[1][i]
+      if not c.isnumeric():
+        variant += c
+      else:
+        break
+      i += 1
+    varSeq = int(pp[1][i]) or 0
+  major = int(v[0])
```
rojer, line 142: more Pythonic way:

major, minor, patch = [int for e in v]

tools/flash_shelly.py
```diff
@@ -0,0 +1,501 @@
...
+
+def write_flash(device, lfw, dlurl, cfw_type, mode):
+  logger.debug("\n" + WHITE + "write_flash" + NC)
+  flashed = False
+  host = device.replace('.local', '')
+  if cfw_type == 'homekit':
+    logger.info("Downloading Firmware...")
+    logger.debug('DURL: %s' % dlurl)
+    myfile = requests.get(dlurl)
+    with open('shelly-flash.zip', 'wb') as f:
```
rojer, line 172: you write this file only to read it back immediately.
if you need a file-like object backed by in-memory data, use io.BytesIO

tools/flash_shelly.py
```diff
@@ -0,0 +1,501 @@
...
+  elif (vi1[3] != vi2[3]):
+    return True
+  elif (vi1[4] != vi2[4]):
+    return (vi1[4] > vi2[4])
+  else:
+    return False
+
+
+def write_flash(device, lfw, dlurl, cfw_type, mode):
+  logger.debug("\n" + WHITE + "write_flash" + NC)
```
rojer, line 165: using + to build strings is not pythonic.
since this is python 3, we can use the `f"..."` operator taht will perform variable interpolation, e.g.

logger.debug(f"\n{WHITE}write_flash{NC}")

much nicer and more readable

rojer (changes requested):

tools/flash_shelly.py
andyblac: doh ye np, don't know what I was thinking there.

tools/flash_shelly.py
andyblac: https://github.com/mongoose-os-apps/shelly-homekit/pull/176/commits/da9181dd0e2545fbb3295e7b47a0cb81921fea92

tools/flash_shelly.py
andyblac: it for update will io.Bytes be ok ?

so...
```
    myfile = requests.get(dlurl)
    with open('shelly-flash.zip', 'wb') as f:
      in_memory = io.BytesIO(f)
    logger.info("Now Flashing...")
    files = {'file': ('shelly-flash.zip', open(in_memory, 'rb'))}
    response = requests.post('http://%s/update' % device , files=files)
    logger.debug(response.text)
```
like that ?

andyblac: done https://github.com/mongoose-os-apps/shelly-homekit/pull/176/commits/58e3935f753894f457811da39e41ebb5146ce7d1

tools/flash_shelly.py
andyblac: done https://github.com/mongoose-os-apps/shelly-homekit/pull/176/commits/ed7b9f018b7e77c98a3700db2e41a9753a03992d

tools/flash_shelly.py
andyblac: hmmm did not work.
```
vi1: (<class 'int'>, <class 'int'>, <class 'int'>, '', 0)
vi2: (<class 'int'>, <class 'int'>, <class 'int'>, '', 0)
```

tools/flash_shelly.py
andyblac: https://github.com/mongoose-os-apps/shelly-homekit/pull/176/commits/758d4c4daeee3692a62120ec21df1d122c2d1736

rojer:
great job!

andyblac:
np, requires python 3.6, I think we should leave both versions up, bash scrpit can be used as fall back, I won't keep it updated with new features, but I'll try to keep basic functionality working.

rojer:
no, i think we should leave just the python version. there isn't any real advantage, just maintenance burden.

andyblac:
ok, np

PR #183, opened 2020-10-09
[FlashScript] recoded to use 'argparse', add --variant, more optimisations, and added more checks.

andyblac:


tools/flash_shelly.py
```diff
@@ -22,19 +22,30 @@
...
+#    -V VERSION, --version VERSION
+#                          Force a particular version.
+#    --variant VARIANT     Prerelease variant name.
+#    -v {0,1}, --verbose {0,1}
+#                          Enable verbose logging level.
 
+
+import argparse
 import functools
 import getopt
```
rojer, line 50: you can remove getopt now

rojer (approved):

tools/flash_shelly.py
andyblac: oops ye

PR #202, opened 2020-10-17
[FlashScript] remove some unnecessary duplicate code.

andyblac:


andyblac:
removed some code, as we have a DICT build when scanning both firmware type are built the same.

PR #211, opened 2020-10-20
[FlashScript] add script to fix 'ssl.SSLCertVerificationError:'

andyblac:


rojer:
hang on, how come curl works and python doesn't? where does curl get its roots?

andyblac:
no idea

rojer:
i'd rather we don't miss with user's SSL trust roots.
these roots should already be on the system, python should be able to find them, especially given that cURL works.
if you wish, you can switch to http:// until you figure it out.

andyblac:
ok, np, there has only been 2 ports of it, maybe it an issue with their systems.

andyblac:
found this https://stackoverflow.com/questions/40684543/how-to-make-python-use-ca-certificates-from-mac-os-truststore

andyblac:
we could just update Wiki pointing this out and say try these method to fis the issue.

In a nutshell: Python 3.6 does not rely on MacOS' openSSL anymore. It comes with its own openSSL bundled and doesn't have access on MacOS' root certificates.

You have two options:

Run an install command shipped with Python 3.6

```
cd /Applications/Python\ 3.6/
./Install\ Certificates.command
```
or

Install the certifi package with

`pip install certifi`

rojer:
can you ask the user to run
```
python3 -c 'import ssl; print(ssl.get_default_verify_paths())'
```
and also
```
ls -l /etc/ssl/
```

andyblac:
i'll use that in wiki

Issue #314, opened 2020-12-02
Set code on first install via cli


timoschilling:
It would be nice if the flash script would accept a `-c/--code` option with direct set the HomeKit code after the install of the Shelly HomeKit firmware.

andyblac:
Please explain more, you mean allow to set HomeKit pairing code ?

timoschilling:
Yes I mean the HomeKit pairing code.

Assume I have an devise with stock firmware, connected to my wifi.

Current workflow:
1. run `flash_shelly.py -y shelly-xxx.local`
2. browse to `http://shelly-xxx.local`
3. enter HomeKit pairing code
4. setup devise in the Home app

Requested workflow:
1. run `flash_shelly.py -y --code 11223344 shelly-xxx.local`
2. setup devise in the Home app

I think doing a `POST http://10.0.0.135/rpc/HAP.Setup` request with `{"code":"112-23-344"}` should do the trick.

andyblac:
sorry but nope
```
Configuring HomeKit setup code...
requests.post(url='http://192.168.3.16/rpc/HAP.Setup', data={'code': '111-22-333'}
{"code":400,"message":"either code or salt + verifier required"}
```

andyblac:
got it done with rojers help

[flash_shelly.py-new-build.zip](https://github.com/mongoose-os-apps/shelly-homekit/files/5632971/flash_shelly.py-new-build.zip)

this build has had a major rewrite, please report any issues.

timoschilling:
Wow that was a fast response / implementation and it works nicely!

But I found one UX glitch, the user needs to use the format `111-22-333` and not `11122333`. If the use uses the wrong format, he didn't get any negative feedback, it only shows: `Configuring HomeKit setup code...`. I would expect a error message or even better that the flash script reformat the number.

And one 💅 point it would be nice to have a handy `-c` or `-h` shortcut for the long `--hap_setup_code`

andyblac:
i'll add a format check and report to user how to format if wrong. i'll also add to help.

rojer:
we should allow 11122333 format, like UI does

andyblac:
ok i'll get script to correct the formating

andyblac:
try this added `-c` too @rojer if you prefer not to have this just say.

[flash_shelly-new-build.zip](https://github.com/mongoose-os-apps/shelly-homekit/files/5636206/flash_shelly-new-build.zip)

[...]

PR #317, opened 2020-12-03
[FlashScript] OOP version, initial commit.

andyblac:


timoschilling:
I assume that this is the PR for this build? https://github.com/mongoose-os-apps/shelly-homekit/issues/314#issuecomment-738239212

If yes I would like to give some feedback:
1. The devise id of stock devises should not contain `.local` as it is by HomeKit devises
2. Maybe you could prefix the host with `http://` that will allow to open the web ui via click from inside the console. In my setup with an unconfigured Safari and an unconfigured iTerm2 terminal it will open all links as `https://` as default. This change will be help to enforce the right `http://` protocol handler.

But at all great work!

andyblac:
didn't know about url links, i don't use iTerm2, and links do not work in default Terminal app.

andyblac:
just tried iterm2 links do not seem to work here.

andyblac:
ahh you have to hold CMD down and click on url

tools/flash-shelly.py
```diff
@@ -0,0 +1,664 @@
...
+  sorted_device_list = sorted(device_list, key=lambda k: k['host'])
+  logger.trace(f"device_test: {sorted_device_list}")
+
+  for device in sorted_device_list:
+    if mode == 'keep':
+      flashmode = device['fw_type']
+    else:
+      flashmode = mode
+
+    if device['fw_type'] == 'homekit':
```
rojer, line 568: nice, i like this part, much better than it used to be!

rojer (commented):
overall looks fine. i haven't done detailed review, but overall seems better than before.

one nitpick: there are three common styles for naming functions: `lower_case_underscore_style()`, `camelCaseStyle()` and `CapitalizedcamelCase()`. amazingly, you managed to use them all :) 
be consistent: pick one and stick to it. fwiw, python style guide recommends `lower_case_underscore_style()`.

tools/flash-shelly.py
andyblac: ok done changed to `lower_case_underscore_style()`

tools/flash-shelly.py
```diff
@@ -0,0 +1,664 @@
...
+    if not self.version:
+      if self.stock_model  == 'SHRGBW2':
+        self.dlurl = stock_model_info['url'].replace('.zip',f'-{self.colour_mode}.zip')
+      else:
+        self.dlurl = stock_model_info['url']
+    else:
+      self.dlurl = f'http://archive.shelly-faq.de/version/v{self.version}/{self.stock_model}.zip'
+
+class HomeKitDevice(Device):
+  def __init__(self, host, wifi_ip, fw_type, device_url, info, variant, version):
```
rojer, line 259: no need to define `__init__` if you're not overriding anything

tools/flash-shelly.py
```diff
@@ -0,0 +1,664 @@
...
+    myfile = requests.get(self.dlurl)
+    logger.info("Now Flashing...")
+    files = {'file': ('shelly-flash.zip', myfile.content)}
+    logger.debug(f"requests.post(f'http://{self.wifi_ip}/update' , files=files")
+    response = requests.post(f'http://{self.wifi_ip}/update' , files=files)
+    logger.debug(response.text)
+
+
+class StockDevice(Device):
+  def __init__(self, host, wifi_ip, fw_type, device_url, info, variant, version):
```
rojer, line 300: same here

tools/flash-shelly.py
```diff
@@ -0,0 +1,664 @@
...
+
+  def add_service(self, zeroconf, type, device):
+    device = device.replace('._http._tcp.local.', '')
+    deviceinfo = Device(device)
+    deviceinfo.get_device_info()
+    if deviceinfo.fw_type is not None:
+      dict = {'host': deviceinfo.host, 'wifi_ip': deviceinfo.wifi_ip, 'fw_type': deviceinfo.fw_type, 'device_url': deviceinfo.device_url, 'info' : deviceinfo.info}
+      self.device_list.append(dict)
+
+class Device():
```
rojer, line 124: remove parens, don't need them

tools/flash-shelly.py
andyblac: done

tools/flash-shelly.py
andyblac: done

tools/flash-shelly.py
andyblac: done

PR #348, opened 2020-12-17
[Flashscript] some tweaks and fixes.

andyblac:
fix possible crash in 'parse_version', Shelly1L has '_1L' in version number.
tweak verbose logging.
allow scanner to timeout if host fails to respond, as this could stop scanner completing a full scan before main scan loop exceeded timeout.

### DIFF
--- a/tools/flash-shelly.py
+++ b/tools/flash-shelly.py
@@ -342,6 +342,8 @@ class StockDevice(Device):
 
 
 def parse_version(vs):
+  # 1.9.2_1L
+  # 1.9.3-rc3 / 2.7.0-beta1
   pp = vs.split('_') if '_' in vs else vs.split('-')
   v = pp[0].split('.')
   variant = ""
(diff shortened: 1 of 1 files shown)


andyblac:
dont pull yet, found another bug.

andyblac:
ok, good to merge.

tools/flash-shelly.py
```diff
@@ -338,8 +342,8 @@ def flash_firmware(self):
 
 
 def parse_version(vs):
```
rojer, line 344: let's collect all the various versions we've seen in the wild and that this function needs tp be able to parse.
this will help with refactoring in the future.

andyblac: upto now i have only seen `2.7.0-beta1` that you use and `1.9.3-rc3` that stock use for variants. and now `1.9.2_1L` fot some strange reason Shelly1L stock release fw uses.

andyblac: we will not know if one off untill 1.9.3 is released.

PR #356, opened 2020-12-23
[FlashScript] use Queue for scanner.

andyblac:
add a Queue for scanner.
add '-q', '--quiet'  to only list devices that need to be upgraded.
add some trace debug info for manual hosts.
add helpful web link to fix online version info failure.
add device name, helps to identify a stock device.
remove the need for urllib and use requests.get.
fix 'friendly_host' variable.
fix manual hosts.
some code cleanup.

### DIFF
--- a/tools/flash-shelly.py
+++ b/tools/flash-shelly.py
@@ -476,7 +476,7 @@ def parse_info(device_info, action, do_selection, dry_run, silent_run, mode, exc
   else:
     latest_fw_label = flash_fw_version
 
-  if do_selection == 'all' or (do_selection == 'upgrade_only' and is_newer(flash_fw_version, current_fw_version)):
+  if do_selection in ('all', 'manual') or (do_selection == 'upgrade_only' and is_newer(flash_fw_version, current_fw_version)):
     global nod
     nod += 1
     logger.info(f"\n{WHITE}Host: {NC}http://{host}")
@@ -530,7 +530,7 @@ def parse_info(device_info, action, do_selection, dry_run, silent_run, mode, exc
         logger.info("Does not need flashing...")
       return 0
 
-    if do_selection == 'all' or (do_selection == 'upgrade_only' and is_newer(flash_fw_version, current_fw_version)):
+    if do_selection in ('all', 'manual') or (do_selection == 'upgrade_only' and is_newer(flash_fw_version, current_fw_version)):
       logger.debug(f"\nperform_flash: {perform_flash}\n")
       if perform_flash == True and dry_run == False and silent_run == False:
         if requires_upgrade == True:
@@ -582,7 +582,7 @@ def probe_device(device, action, do_selection, dry_run, silent_run, mode, exclud
 
 def device_scan(hosts, action, do_selection, dry_run, silent_run, mode, exclude, version, variant, hap_setup_code):
   logger.debug(f"\n{WHITE}device_scan{NC}")
-  if not do_selection:
+  if hosts:
     for host in hosts:
       deviceinfo = Device(host)
       deviceinfo.get_device_info()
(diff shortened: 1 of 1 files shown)


tools/flash-shelly.py
```diff
@@ -601,13 +565,51 @@ def device_scan(hosts, action, do_all, dry_run, silent_run, mode, exclude, versi
...
+    listener = MyListener()
+    browser = zeroconf.ServiceBrowser(zc, '_http._tcp.local.', listener)
+    zc.wait(100)
+    time.sleep(1)
+    d_queue = listener.queue
+    total_loop = 1
+    nod = 0
+    scan_finished = zc.done
+    while not scan_finished:
+      while d_queue.empty() and total_loop < 6: # do 6 loops of 2 seconds, if queue is still empty close scanner.
```
rojer, line 598: use queue.get with timeout of, like, 20 seconds. when it times out, we're done

andyblac: sorry i dont understand, how does timeout help with looping queue ?, and detecting when queue is empty ?

andyblac: never mind, think I understand.

tools/flash-shelly.py
```diff
@@ -601,18 +582,51 @@ def device_scan(hosts, action, do_all, dry_run, silent_run, mode, exclude, versi
...
+    while not scan_finished:
+      try:
+        device = d_queue.get(timeout=20)
+        d_info = json.dumps(device, indent = 4)
+        logger.trace(f"Device Info: {d_info}")
+        probe_device(device, action, do_selection, dry_run, silent_run, mode, exclude, version, variant, hap_setup_code)
+      except queue.Empty:
+        logger.info(f"\n{GREEN}Devices found: {nod}{NC}")
+        zc.close()
+        scan_finished = zc.done
```
rojer, line 622: from what i can tell, zc.close() just sets zc.done to true, i don't think we need to check it explicitly.

```
while True:
...
  except queue.Empty:
    zc.close()
    break
 ```

File: tools/flash-shelly.py [hunk 3]
```diff
@@ -601,18 +582,51 @@ def device_scan(hosts, action, do_all, dry_run, silent_run, mode, exclude, versi
...
+        device = {'host': deviceinfo.host, 'wifi_ip': deviceinfo.wifi_ip, 'fw_type': deviceinfo.fw_type, 'device_url': deviceinfo.device_url, 'info': deviceinfo.info}
+        d_info = json.dumps(device, indent = 4)
+        logger.trace(f"Device Info: {d_info}")
+        probe_device(device, action, do_selection, dry_run, silent_run, mode, exclude, version, variant, hap_setup_code)
+  else:
+    logger.info(f"{WHITE}Scanning for Shelly devices...{NC}")
+    zc = zeroconf.Zeroconf()
+    listener = MyListener()
+    browser = zeroconf.ServiceBrowser(zc, '_http._tcp.local.', listener)
+    zc.wait(100)
```
rojer, line 608: no need to wait here

tools/flash-shelly.py
```diff
@@ -601,18 +582,51 @@ def device_scan(hosts, action, do_all, dry_run, silent_run, mode, exclude, versi
...
+    time.sleep(1)
```
rojer, line 609: ... or to sleep

tools/flash-shelly.py
```diff
@@ -601,18 +582,51 @@ def device_scan(hosts, action, do_all, dry_run, silent_run, mode, exclude, versi
...
+    d_queue = listener.queue
```
rojer, line 610: don't need this var either

tools/flash-shelly.py
andyblac: the wait i found is required, it is a timeout for browser to return a result, else it can stall and lockup.

```
wait(timeout: float) → None
Calling thread waits for a given number of milliseconds or until notified.
```

tools/flash-shelly.py
andyblac: the slepp was required as it took a seconds to get 1st item added to queue, butthis was when i was checking against empty queue, so yes i can remove now.

tools/flash-shelly.py
andyblac: ok yes, no longer require it.

tools/flash-shelly.py
andyblac: by this i mean, if a device is online and has an IP, but for some reason it fails to respond to requests, like in my current situation, it just sits there waiting , and waiting, with a 100ms timeout, it silently just ignores the device.

rojer: missing devices is possible but calling or not calling Zeroconf.wait() has nothing to do with it.
looking at the implementation, it just waits on an internal condition variable.
all the scanning is done in background threads

andyblac:
ok ready for another review.

PR #380, opened 2021-01-03
Make Web UI more mobile friendly

ZeevoX:
* Support screens as narrow as 300px
* Responsively adapt on rotation to/from landscape mode
* Properly wrap text when it is necessary to do so

#### How:
* Turned each section into an HTML table: replaced the `div`s with `table`s and `tr`s, wrapped the `label`s and values in `tr`s. 
* CSS media query checking for screen width
* NB: had to replace `block` with `table-row` in the JS so that the rows would show up correctly for programmatically-enabled elements

<details>
<summary>Smartphone screenshot demo (click me)</summary>

</details>


andyblac:
is this getting redone?, any reason why it was closed ?

PR #396, opened 2021-01-09
[FlashScript] Logging tweaks

andyblac:
add file logging via --log-file {LOG_FILENAME}
add --verbose levels 0 - 5
add 'flashed_devices' count

### DIFF
--- a/tools/flash-shelly.py
+++ b/tools/flash-shelly.py
@@ -158,9 +158,9 @@ class Device:
         if fp.status_code == 200:
           info = json.loads(fp.content)
       except requests.exceptions.RequestException as err:
-        logger.critical(f"Error: {err}")
+        logger.debug(f"Error: {err}")
     else:
-      logger.warning(f"Could not get info from device: {self.host}")
+      logger.debug(f"Could not get info from device: {self.host}")
     self.info = info
     return info
 
--- a/tools/flash-shelly.py
+++ b/tools/flash-shelly.py
@@ -69,6 +69,15 @@ logging.Logger.trace = functools.partialmethod(logging.Logger.log, logging.TRACE
 logging.trace = functools.partial(logging.log, logging.TRACE)
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.TRACE)
+def log_level(level):
+  options = {'0' : logging.CRITICAL,
+             '1' : logging.ERROR,
+             '2' : logging.WARNING,
+             '3' : logging.INFO,
+             '4' : logging.DEBUG,
+             '5' : logging.TRACE,
+   }
+  return options.get(level, level)
 
 upgradeable_devices = 0
 flashed_devices = 0
@@ -475,6 +484,7 @@ def parse_info(device_info, action, dry_run, quiet_run, silent_run, mode, exclud
     latest_fw_label = flash_fw_version
 
   if (not quiet_run or (quiet_run and (is_newer(flash_fw_version, current_fw_version) or force_version and dlurl and parse_version(flash_fw_version) != parse_version(current_fw_version)))) and requires_upgrade != 'Done':
+    logger.info(f"")
     logger.info(f"{WHITE}Host: {NC}http://{host}")
     logger.info(f"{WHITE}Device Name: {NC}{device_name}")
     logger.info(f"{WHITE}Device ID: {NC}{device_id}")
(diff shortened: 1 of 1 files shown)


rojer:
traditionally -v controls the _level_ of verbosity, the higher the more verbose.
logging to file can easily be done with input redirection, so i don't see the need, as such.
so, how about you make -v control the level (instead of just fixing it to DEBUG)

andyblac:
> traditionally -v controls the _level_ of verbosity, the higher the more verbose.
> logging to file can easily be done with input redirection, so i don't see the need, as such.
> so, how about you make -v control the level (instead of just fixing it to DEBUG)

ok, v=0 just list normal INFO, and -v=1 set to DEBUG, and only output to file and screen if -v=1 ?

andyblac:
and default would be -v=0 if nothing is set in command line.

andyblac:
> logging to file can easily be done with input redirection, so i don't see the need, as such.

how do you propose to do it, make it easy for users ?

andyblac:
how about `--log` can be used with or without `-v=?`

rojer:
how about `--log-file=FILE_NAME`

andyblac:
that  ok ?

andyblac:
hmm needs work

andyblac:
ok try that, look ok ?

tools/flash-shelly.py
```diff
@@ -644,24 +629,55 @@ def device_scan(hosts, action, dry_run, quiet_run, silent_run, mode, exclude, ve
...
+    fh = logging.FileHandler(args.log_filename, mode='w', encoding='utf-8')
+    fh.setFormatter(logging.Formatter('%(asctime)s %(levelname)s %(lineno)d %(message)s'))
+    if args.verbose == '0':
+      fh.setLevel(logging.INFO)
+    elif args.verbose == '1':
+      fh.setLevel(logging.DEBUG)
+    logger.addHandler(sh)
+    logger.addHandler(fh)
+  else:
+    logger.addHandler(sh)
```
rojer, line 654: logger.addHandler(sh) is in both branches, can be moved up a level

tools/flash-shelly.py
```diff
@@ -45,7 +45,7 @@
 #                          Configure HomeKit setup code, after flashing.
 #    --variant VARIANT     Prerelease variant name.
 #    -v {0,1}, --verbose {0,1}
-#                          Enable verbose logging level.
+#                          Enable verbose logging 0=screen / 1=file.
```
rojer, line 48: this is not correct anynmore

tools/flash-shelly.py
```diff
@@ -644,24 +629,55 @@ def device_scan(hosts, action, dry_run, quiet_run, silent_run, mode, exclude, ve
   parser.add_argument('-y', '--assume-yes', action="store_true", dest='silent_run', default=False, help="Do not ask any confirmation to perform the flash.")
   parser.add_argument('-V', '--version',type=str, action="store", dest="version", default=False, help="Force a particular version.")
   parser.add_argument('-c', '--hap-setup-code', action="store", dest="hap_setup_code", default=False, help="Configure HomeKit setup code, after flashing.")
+  parser.add_argument('-v', '--verbose', action="store", dest="verbose", choices=['0', '1'], default='0', help="Enable verbose logging 0=info / 1=debug.")
```
rojer, line 632: 0 is not a good default because there should be a way to decrease verbosity as well as increase.
decreasing makes sense if user want no output - makes sense with `-a -y`, for exampel - just upgrade evrything and don't bother me.
so, i suggest we have a level of verbosity for each of the logging levels - trace, debug, info, warning, error, critical (5, ,4, 3, 2, 1, 0 respectively) and set default to 3 (info).

tools/flash-shelly.py
andyblac: ok

tools/flash-shelly.py
andyblac: ok done

tools/flash-shelly.py
andyblac: done

tools/flash-shelly.py
andyblac: ok done.

tools/flash-shelly.py
```diff
@@ -644,24 +640,65 @@ def device_scan(hosts, action, dry_run, quiet_run, silent_run, mode, exclude, ve
...
   action = 'list' if args.list else 'flash'
   args.mode = 'stock' if args.mode == 'revert' else args.mode
   args.hap_setup_code = f"{args.hap_setup_code[:3]}-{args.hap_setup_code[3:-3]}-{args.hap_setup_code[5:]}" if args.hap_setup_code and '-' not in args.hap_setup_code else args.hap_setup_code
-  if args.verbose and '0' in args.verbose:
-    logger.setLevel(logging.DEBUG)
-  elif args.verbose and '1' in args.verbose:
-    logger.setLevel(logging.TRACE)
+  sh = logging.StreamHandler()
+  sh.setFormatter(logging.Formatter('%(message)s'))
+  if args.verbose == '0':
```
rojer, line 653: this kind of mapping is best done with a map

tools/flash-shelly.py
```diff
@@ -644,24 +640,65 @@ def device_scan(hosts, action, dry_run, quiet_run, silent_run, mode, exclude, ve
...
+  elif args.verbose == '3':
+    sh.setLevel(logging.INFO)
+  elif args.verbose == '4':
+    sh.setLevel(logging.DEBUG)
+  elif args.verbose == '5':
+    sh.setLevel(logging.TRACE)
+  if args.log_filename:
+    fh = logging.FileHandler(args.log_filename, mode='w', encoding='utf-8')
+    fh.setFormatter(logging.Formatter('%(asctime)s %(levelname)s %(lineno)d %(message)s'))
+    if args.verbose in ('1', '2', '3'):
```
rojer, line 668: why is this different for file logging vs console? they should be the same.

andyblac: warning etc only in log, means no sence to me.

tools/flash-shelly.py
andyblac: how do you mean ?

tools/flash-shelly.py
rojer: if we want to have separate log levels for file and console, we can add a separate flag.
but i think -v should just apply to both the same.

tools/flash-shelly.py
andyblac: like this https://github.com/mongoose-os-apps/shelly-homekit/pull/396/commits/0568bef53ef4e7cff8434cbbe6bb3b9bb247301f

tools/flash-shelly.py
andyblac: ok

tools/flash-shelly.py
```diff
@@ -64,29 +67,21 @@
 logging.addLevelName(logging.TRACE, 'TRACE')
 logging.Logger.trace = functools.partialmethod(logging.Logger.log, logging.TRACE)
 logging.trace = functools.partial(logging.log, logging.TRACE)
-logging.basicConfig(format='%(message)s')
 logger = logging.getLogger(__name__)
-logger.setLevel(logging.INFO)
+logger.setLevel(logging.TRACE)
+def log_level(level):
```
rojer, line 72: this doesn't have to be a function, make it a map variable.

LOG_LEVEL = {
  '0' : logging.CRITICAL,
  ...
}


...

sh.setLevel(LOG_LEVEL[args.verbose])

andyblac: ah ok

andyblac: done.

Issue #342, opened 2020-12-15
Led disable to current savings/lowering of temperature...


lucaspinelli85:
Yes, it is very little, perhaps irrelevant, but in original firmware there is an option to disable the led status.
Is it simple to integrate it in webui?
The devive is locked up, not visible from the outside, it is useless to have the led on which is not needed...

rojer:
the default is to have LED off if device is connected to wifi and paired with a HAP controller, so it should be off for most of the time
is that not the case?

lucaspinelli85:
You are right, sorry!

lucaspinelli85:
Not in Shelly PlugS, blue led blinking, red led on when power on...

rojer:
ok, you're right. could be done, but it's so minor i don't know if it even deserves an option.
the amount of current drawn by the LED is miniscule. there may be an argument about light pollution but... meh, i don't know.
i'll let this one simmer and gather support. if you are reading this and want this to happen, please speak up.

lucaspinelli85:
it's just the hassle of always seeing the LED lit in the plug even when it is not active ...

pivale:
It might be a problem for people with one in a bedroom during the night.
Some people have trouble sleeping with the faintest light...

chstoll:
It would be nice to get an option to disable the red LED of the PlugS. In some of my scenes the red LED is just a little disturbing, So +1 ...

timoschilling:
+1

Spy-1:
+1

PR #415, opened 2021-01-19
make power state led configurable on PlugS

timoschilling:
Make the power state led of the PlugS configurable, closes https://github.com/mongoose-os-apps/shelly-homekit/issues/342.

Current testet on a PlugS, but maybe we can expand that pattern to even more shelly types.


rojer:
thanks Timo! this looks good but we need this option to not be displayed at all if there is no LED to speak of (i.e. if led_out_ is null).

timoschilling:
Yes you are right, I noticed it by my self, I'm working on a fix.
I switched `power_state_led` from an bool to an int:
- -1 means UI didn't shown
- 0 means LED off
- 1 means LED on

timoschilling:
updated

mos.yml
```diff
@@ -48,6 +48,7 @@ config_schema:
   - ["sw.initial_state", "i", 3, {title: "Initial state on power-on: 0 - off, 1 - on, 2 - restore last state, 3 - matches input if in toggle mode, otherwise off"}]
   - ["sw.auto_off", "b", false, {title: "Whether the switch should automatically turn OFF after turning ON"}]
   - ["sw.auto_off_delay", "d", 0, {title: "Delay for automatically turning OFF, in seconds"}]
+  - ["sw.power_state_led", "i", -1, {title: "Power state LED enabled"}]
```
rojer, line 51: explain that `-1` means no power led.

rojer (commented):
pretty good, just a minor comment. plus resolve the index.html.gz conflict and we're good to go.

mos.yml
```diff
@@ -48,7 +48,7 @@ config_schema:
   - ["sw.initial_state", "i", 3, {title: "Initial state on power-on: 0 - off, 1 - on, 2 - restore last state, 3 - matches input if in toggle mode, otherwise off"}]
   - ["sw.auto_off", "b", false, {title: "Whether the switch should automatically turn OFF after turning ON"}]
   - ["sw.auto_off_delay", "d", 0, {title: "Delay for automatically turning OFF, in seconds"}]
-  - ["sw.power_state_led", "i", -1, {title: "Power state LED enabled"}]
+  - ["sw.power_state_led", "i", -1, {title: "Power state LED: -1 - unsupported by device, 0 - off, 1 - on"}]
```
rojer, line 51: let's call this `sw.state_led_en`. -1 - unsupported, 0 - disabled, 1 - enabled

please update other places accordingly

fs_src/index.html
```diff
@@ -282,6 +282,13 @@ <h1 id="head">Switch</h1>
         <input type="text" id="auto_off_delay" placeholder="D:HH:MM:SS.sss" required
                pattern="[0-9]+:(0[0-9]|1[0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9]\.[0-9]{3}">
       </div>
+      <div class="form-control" id="power_state_led_container">
+        <label for="power_state_led">Power state led:</label>
```
rojer, line 286: State LED

rojer (commented):
almost there

mos.yml
timoschilling: I can do, I only want to say that there are different status LEDs in the Shelly's. Power in the PlugS, Connect/Paired in the PlugS, Wifi in the Dimmer/RGBW. So I think we should be more precise with the naming.

rojer: but this one will be under specific switch section, i think it's clear. we can have a setting under system settings for system led indication.

timoschilling:
@rojer it’s updated

src/shelly_switch.cpp
```diff
@@ -138,6 +139,10 @@ Status ShellySwitch::SetConfig(const std::string &config_json,
   cfg_->initial_state = cfg.initial_state;
   cfg_->auto_off = cfg.auto_off;
   cfg_->auto_off_delay = cfg.auto_off_delay;
+  if (cfg_->state_led_en != cfg.state_led_en) {
```
rojer, line 142: i think we should disallow setting to any value if it's currently -1.

timoschilling: Like this?
```suggestion
  if (cfg_->state_led_en != -1 && cfg_->state_led_en != cfg.state_led_en) {
```

rojer: even earlier, during validation, we should check for valid new values - 0, 1 and that previous not -1, and send an error

timoschilling: would you like add append an commit with a change according to your ideas

rojer: i don't think i can push directly to this branch, but here's what i meant:

```
  if ((cfg_->state_led_en == -1 && cfg.state_led_en != -1) ||
      (cfg_->state_led_en != -1 && cfg.state_led_en != 0 &&
       cfg.state_led_en != 1)) {
    return mgos::Errorf(STATUS_INVALID_ARGUMENT, "invalid %s", "state_led_en");
  }
```

src/shelly_switch.cpp
```diff
@@ -115,6 +116,9 @@ Status ShellySwitch::SetConfig(const std::string &config_json,
   if (cfg.initial_state < 0 || cfg.initial_state > (int) InitialState::kMax) {
     return mgos::Errorf(STATUS_INVALID_ARGUMENT, "invalid %s", "initial_state");
   }
+  if (cfg.state_led_en < 0 || cfg.state_led_en > 1) {
+    return mgos::Errorf(STATUS_INVALID_ARGUMENT, "invalid %s", "state_led_en");
+  }
```
timoschilling, line 121: @rojer like this?

rojer: not quite, see my comment above

src/shelly_switch.cpp
timoschilling: updated my branch

rojer:
merged, thank you!

timoschilling:
🎉 nice to be part of this great project!!!

eSportler89:
Good morning, with Version _2.9.1_, Build _20210518-010122/2.9.1-gb7dcbef-master_ I got the problem of a blinking blue LED. Changing the Powerstate LED works fine, but is there a way to disable the wifi status light?
Normally it would be solid when connected and blinking when it's in AP mode.

Thanks!

timoschilling:
The blue LED is off when connected to a WiFi and connected to HomeKit

eSportler89:
Argh, damn!
Configured it, connected to my wifi but didn't add it to my Homekit.... I only tested it within the webinterface.
Thank you! :)

PR #417, opened 2021-01-20
Fix UI bug with selected text in inputs (fix #392)

### LINKED ISSUE #392: UI bug with selecting text in inputs
Way to reproduce:
1. Open UI of a devise (tested with PlugS, i3 and 2.5)
2. Select the text of a text input for example devise name
3. wait one or two second and the text will be deselected

This happens until the first change in the inputs happens, after that selected text keeps selected.

ZeevoX:
* Do not update elements if there is an active selection and/or their
  value has not changed since the last refresh


fs_src/script.js
```diff
@@ -846,12 +846,21 @@ function refreshUI() {
 }
 
 function setValueIfNotModified(e, newValue) {
-  if (e.lastSetValue !== undefined && e.lastSetValue !== e.value) return;
+  // do not update the value of the input field if
+  if (e.selected ||                    // the user has selected / highlighted the input field OR
+    e.lastSetValue === e.value ||      // the value has not been changed by the user OR
+    e.lastSetValue !== undefined       // a value has previously been set AND
```
rojer, line 852: please put parens around && when combined with ||: `a || (b && c)` so reader does not need to remember operator precedence rules in mind when reading.

also, please leave operator at the end:

```
  a ||
  (b &&
   c)
```

rojer:
thank you!

pro tip: if you put "Closes #XXX" or "Fixes #XXX" in your commit message (as opposed to subject), issue XXX will be closed automatically.

PR #418, opened 2021-01-20
Link update alert to the system update section (Fix #303)

### LINKED ISSUE #303: Make the Update Alert a link to the System.Update section
Today I clicked the **Update** Alert icon on the top of the Shelly1 internal webpage expecting it to be a button and it to start the update process or take me to the update section.

It doesn’t.

But then I thought: it looks like a button and I clicked it because I though it was a button.

So why not make it a button, or a clickable link to take you to the Update line in the System section?
That should be easily doable with relative URL linking within the same page...

ZeevoX:
* use `div`'s `onclick` to teleport user to correct section
* style the mouse pointer so that it is clear that it is clickable

Screenshot from 2021-01-20 19-24-28

### DIFF
--- a/fs_src/index.html
+++ b/fs_src/index.html
@@ -12,18 +12,19 @@
 </head>
 <body onLoad="onLoad()">
 
-<div class="container">
+<div class="container" id="header_container">
   <a class="helpbadge" href="https://github.com/mongoose-os-apps/shelly-homekit/wiki/" target="_blank">?</a>
-  <div class="logodiv"><img class="img" src="logo.png" alt=""></div>
-  <div class="apl_thin">
-    Shelly-<b>HomeKit</b>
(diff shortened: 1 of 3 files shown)


rojer:
instead of this, can we make badges <a>s instead of <div>s? this way we can use a's href attribute.

ZeevoX:
sorry, I don't understand what you mean by s

rojer:
oh, github butchered the html tags in the comment. fixed.

andyblac:
linked to issue https://github.com/mongoose-os-apps/shelly-homekit/issues/303

ZeevoX:
image

re-did it using `<a>` tags. also stopped badges overlaying on top of each other if shown simultaneously.

rojer:
nice! please rebase and it's good to go

rojer:
thank you!

PR #420, opened 2021-01-20
git ignore fs/index.html.gz

timoschilling:
`fs/index.html.gz` is an generated binary file so it should be ignored in the repository.


rojer:
i want people who cannot regenerate it (e.g. because they are on windows) to still be able to build the firmware, that's why a prebuilt version is provided.

timoschilling:
I understand your argument.

How does the build work on windows? On the "Windows-Subsytem for Linux" or on Docker or some other way?
If it's one of the both first once, it should be possible to create it on windows as well.

rojer:
i am not using windows myself but it should be possible with just having mos installed and using remote build, so no docker  necessary.

PR #425, opened 2021-01-22
[FlashScript] version 2.1.5

andyblac:
only show Sys Temp if key exists
allow more time for device to flash and reboot, some stock device's can be slow to perform a flash.
add device count for flash failed devices, only show flashed / failed when in flashing mode.
fix varSeq in version parse.
fix forced version flashing for stock firmware.

### DIFF
--- a/tools/flash-shelly.py
+++ b/tools/flash-shelly.py
@@ -352,7 +352,7 @@ def parse_version(vs):
       else:
         break
       i += 1
-    varSeq = int(pp[1][i]) if len(pp[1]) >= i else 0
+    varSeq = int(pp[1][i]) if len(pp[1]) > i else 0
   major, minor, patch = [int(e) for e in v]
   return (major, minor, patch, variant, varSeq)
 
--- a/tools/flash-shelly.py
+++ b/tools/flash-shelly.py
@@ -478,7 +478,7 @@ def parse_info(device_info, action, dry_run, quiet_run, silent_run, mode, exclud
     latest_fw_label = f"{RED}No {device_info.variant} available{NC}"
     flash_fw_version = '0.0.0'
     dlurl = None
-  elif not dlurl or durl_request.status_code != 200:
+  elif not dlurl or durl_request.status_code != 200 or ('Content-Type' in durl_request.headers and durl_request.headers['Content-Type'] != 'application/zip'):
     latest_fw_label = f"{RED}Not available{NC}"
     flash_fw_version = '0.0.0'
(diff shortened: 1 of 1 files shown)


tools/flash-shelly.py
```diff
@@ -473,7 +478,7 @@ def parse_info(device_info, action, dry_run, quiet_run, silent_run, mode, exclud
     latest_fw_label = f"{RED}No {device_info.variant} available{NC}"
     flash_fw_version = '0.0.0'
     dlurl = None
-  elif not dlurl or durl_request.status_code != 200:
+  elif not dlurl or durl_request.status_code != 200 or ('Content-Type' in durl_request.headers and durl_request.headers['Content-Type'] != 'application/zip'):
```
rojer, line 481: durl_request.headers.get('Content-Type', '') != 'application/zip'

tools/flash-shelly.py
```diff
@@ -443,7 +448,7 @@ def parse_info(device_info, action, dry_run, quiet_run, silent_run, mode, exclud
   colour_mode = device_info.colour_mode
   dlurl = device_info.dlurl
   flash_label = device_info.flash_label
-  sys_temp = device_info.info['sys_temp'] if 'sys_temp' in device_info.info else '0'
+  sys_temp = device_info.info['sys_temp'] if 'sys_temp' in device_info.info else None
```
rojer, line 451: sys_temp = device_info.info.get('sys_temp', None)

tools/flash-shelly.py
```diff
@@ -487,7 +492,8 @@ def parse_info(device_info, action, dry_run, quiet_run, silent_run, mode, exclud
     logger.info(f"{WHITE}Device ID: {NC}{device_id}")
     logger.info(f"{WHITE}IP: {NC}{wifi_ip}")
     logger.info(f"{WHITE}Model: {NC}{model}")
-    logger.info(f"{WHITE}Sys Temp: {NC}{sys_temp}˚c{NC}")
+    if sys_temp:
```
rojer, line 495: if sys_temp is not None

it's unlikely , but we could have system temperature of 0 degrees

andyblac: done

tools/flash-shelly.py
andyblac: done

tools/flash-shelly.py
andyblac: done

Issue #43, opened 2020-04-08
Water valve support


Snekatis:
Is it possible add support for Water valve? So HomeKit will see it as a real Water valve?

![HK](https://user-images.githubusercontent.com/459219/78779622-87990400-799d-11ea-83db-a9235393201c.png)

ursusHU:
And also Garage door pleasae :)

Montardit:
Yes, please! Sprinkles support!

danny1605:
Please, integrate the irrigation type !

lucaspinelli85:
it would be great for my garden, please integrate it!

andyblac:
pull request here https://github.com/mongoose-os-apps/shelly-homekit/pull/433

danny1605:
There is already new information regarding Water valve support ?

timoschilling:
Basic valve support is now merged, timer support maybe follow.

rojer:
please help test at https://github.com/mongoose-os-apps/shelly-homekit/issues/498

Montardit:
Shellys configured as water valve can't be added to automatizations. If I configure it as as switch it is selectable but if it is configured as a valve not. Not as a irrigation nor as basic valve. I've configured it with a shelly 2.5. They don't show in the accessories/ambient selector list of the automatization.

timoschilling:
@Montardit I can confirm that, but it looks like a bug or missing feature in the home app. I created an irrigation valve with the HomeKit Accessory Simulator (a developer app from Apple) and it behaves the same. Google for that and you will find some more reports about that. But you can use apps like Home+ or Eve to create a irrigation related automatisation, maybe using a scene to start the irrigation from a automatisation.

Montardit:
It's a little weird, I hope they'll fix this. It's strange you cannot 
water your plants regularly if you have an irrigation valve... I've 
tried with Eve and it works fine. Home+ is not free.

Thanks!!!

El 17/3/21 a les 3:50, Timo Schilling ha escrit:

danny1605:
hello, is there anything new regarding the timing of irrigation?

blitzi2000:
maybe adding a new HAP Type in the roller shutter mode so can be seeing as a water tap/sprinkler in homekit? that would allow opening and closing and automatize with a time setting...

PR #433, opened 2021-01-25
Valve

timoschilling:
Adding valve as a switch type, including support for the different valve types. Timer support will be maybe added later.


andyblac:
wiki will need updating too once comittied.

andyblac:
link to issue https://github.com/mongoose-os-apps/shelly-homekit/issues/43

fs_src/index.html
```diff
@@ -236,12 +236,22 @@ <h1 id="head">Switch</h1>
...
+        <label>HAP Service Type:</label>
         <select id="svc_type">
           <option id="svc_type_-1" value="-1">Disabled</option>
           <option id="svc_type_0" value="0">Switch</option>
           <option id="svc_type_1" value="1">Outlet</option>
           <option id="svc_type_2" value="2">Lock</option>
+          <option id="svc_type_3" value="3">Valve</option>
+        </select>
+      </div>
+      <div class="form-control" id="char_type_container" style="display: none">
```
rojer, line 248: it's valve type, let's call it that.

src/shelly_hap_valve.cpp
```diff
@@ -0,0 +1,102 @@
...
+      std::bind(&Valve::HandleValveTypeRead, this, _1, _2, _3),
+      true /* supports_notification */, nullptr /* write_handler */,
+      kHAPCharacteristicDebugDescription_ValveType);
+  state_notify_chars_.push_back(valve_type_char);
+  AddChar(valve_type_char);
+
+  return Status::OK();
+}
+
+HAPError Valve::HandleActiveRead(
```
rojer, line 73: these are simple enough to be lambdas

src/shelly_hap_valve.hpp
```diff
@@ -0,0 +1,47 @@
...
+namespace shelly {
+namespace hap {
+
+class Valve : public ShellySwitch {
+ public:
+  Valve(int id, Input *in, Output *out, PowerMeter *out_pm, Output *led_out,
+       struct mgos_config_sw *cfg);
+  virtual ~Valve();
+
+  Status Init();
```
rojer, line 32: Status Init() override;

rojer (commented):
hey, not bad! couple comments and this is good to go.

fs_src/index.html
timoschilling: I chose the generic name to make them reusable for other cases as well.
- Security System Alarm Type
- Slat Type

rojer: they will have different nomenclature though. we'll use different parameter for these, if and when they are implemented.

src/shelly_hap_valve.cpp
timoschilling: I followed the style from the shelly hap lock

src/shelly_common.hpp
```diff
@@ -51,6 +52,8 @@
 #define SHELLY_HAP_IID_BASE_OCCUPANCY_SENSOR 0x800
 #define SHELLY_HAP_IID_BASE_CONTACT_SENSOR 0x900
 #define SHELLY_HAP_IID_STEP_SENSOR 0x10
+#define SHELLY_HAP_IID_BASE_VALVE 0xa00
+#define SHELLY_HAP_IID_STEP_VALVE 4
```
rojer, line 56: valve can have more characteristics. let's leave some space, just in case we ever need it. set step to 0x10

src/shelly_hap_valve.cpp
rojer: ok, i'll refactor it myself, no problem.

rojer (commented):
almost there, just a minor thing

rojer:
thank you!

danny1605:
when will the update be available? so that I can use the valve? thank you in advance for inserting the valve option. Hope that a timer function will come

rojer:
@danny it will be in 2.8.0. release cycle has not been started yet, but i think we'll cut in the next couple weeks.

aae55555:
did you guys ever manage to figure out the timer function?

timoschilling:
There was an unresolved problem with the timer, so it's still not implemented.

PR #440, opened 2021-01-27
add confirm step to some UI actions

timoschilling:
Actual a click on "revert" (stock firmware), "reset" (devise settings / hap settings) and "reboot" will be directly executed. I think that's not a good UI, unrevertible action should be confirmed by the user. I added the confirmations in this PR.


andyblac (commented):
Should this not be asking "Are you sure you want this device to be reverted to stock firmware?", this sounds better English IMO.

timoschilling:
I'm ok with any text.

timoschilling:
I'm ok with any text, would you like to provide a text for each case?

andyblac:
Maybe @rojer has a preference?

fs_src/script.js
```diff
@@ -340,12 +342,16 @@ function gdoSetConfig(c, cfg, spinner) {
...
 el("reboot_btn").onclick = function () {
+  if(!confirm("Should this device perform a reboot?")) return;
+
   sendMessageWebSocket("Sys.Reboot", {delay_ms: 500}).then(function () {
     alert("System is rebooting and will reconnect when ready.");
   });
 }
 
 el("reset_btn").onclick = function () {
+  if(!confirm("Should this device perform a reset?")) return;
```
rojer, line 353: "Device configuration will be wiped and return to AP mode. Are you sure?"

fs_src/script.js
```diff
@@ -1040,6 +1046,8 @@ el("update_btn").onclick = function () {
   checkUpdate();
 };
 el("revert_btn").onclick = function () {
+  if(!confirm("Should this device perform a firmware reset?")) return;
```
rojer, line 1049: "Revert to stock firmware?"

PR #447, opened 2021-01-28
replace logo.png with a svg

timoschilling:
This reduces the filesystem content by ~4.4kb.

I also tried to use a base64 encoded version of the `logo.png`, but that increases the size of `index.html.gz` by ~5k, which means no improvement at all.

The SVG is hand coded, so it doesn't match the old logo 1:1, but I think it's ok, maybe someone can improve it in the future.

I also would like to use that svg as favicon, but Safari didn't support that at the current version.


rojer:
nice! please add logo.png to the list of obsolete files [here](https://github.com/mongoose-os-apps/shelly-homekit/blob/7da43a25513946e59a178a56540fed3dd3b47e0f/src/shelly_main.cpp#L786) so it doesn't get copied during OTA

rojer:
thank you!

PR #449, opened 2021-01-28
fixed #404 adding manual calibration.

### LINKED ISSUE #404: use shelly 2.5 roller shutter without calibration
I'm trying to use the shelly 2.5 with a **Velux KLF 050**. The Velux KLF 050 is a remote for the Velux Solar roller shutter and provides inputs for the opening and closing.

when using the original firmware I can open and close the roller shutter by providing an opening/closing time of about 55 seconds.

Unfortunately it is not possible to use calibration because the KLF 050 **won't consume any power** unlike a roller shutter engine, so calibration will always fail.

With the homekit-firmware opening/closing is not possible - and i even cannot set the time taken for opening/closing.
and of course opening to 50% isn't possible too because of missing calibration.

without calibration the shelly cannot be used I understand.

Is it somehow possible to circumvent the calibration process and provide the needed data by setting some configuration manually like opening/closing time or calibration data?

juangamnik:
I added the ability to manually calibrate roller shutters (window closing). This should be used for roller shutters with end position detection, only (should be the case for any motor newer than 25 years). Therefore, a "Caution" hint is added as soon as manual calibration is activated. Two values can be set: (1) the move time. (2) For movements to limit positions a second (i.e., longer) movement time (to limit pos) can be added.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ deps/
 releases/
 .DS_Store
 *.zip
+.vscode/
--- a/fs_src/index.html
+++ b/fs_src/index.html
@@ -457,6 +457,27 @@
         <span id="cal"></span>
       </div>
       <div class="form-control">
+        <label>Manual Calibration:</label>
+        <label class="switch">
+          <input type="checkbox" id="man_cal">
+          <span class="slider round"></span>
+        </label> 
+        <div id="man_cal_warning">
+          <ion-icon name="warning" 
+            style="vertical-align: text-bottom; font-size: 24px; color: goldenrod;"></ion-icon> Caution
+        </div>
+      </div>
+      <div id="man_cal_settings">
+        <div class="form-control">
+          <label>Movement Time:</label>
+          <input type="text" id="move_time_ms" style="width: 4em; min-width: 2em;"> ms
+        </div>
+        <div class="form-control">
+          <label>Limit Time:</label>
+          <input type="text" id="move_time_limit_pos_ms" style="width: 4em; min-width: 2em;"> ms
+        </div>
+      </div>
+      <div class="form-control">
         <label>Input Mode:</label>
         <select id="in_mode">
           <option id="in_mode_0" value="0">Separate - momentary</option>
@@ -552,5 +573,6 @@
 <script>
 
 </script>
+<script src="https://unpkg.com/ionicons@5.0.0/dist/ionicons.js"></script>
 </body>
 </html>
--- a/fs_src/script.js
+++ b/fs_src/script.js
@@ -295,6 +295,9 @@ function wcSetConfig(c, cfg, spinner) {
     cfg = {
       name: name,
       in_mode: parseInt(el(c, "in_mode").value),
+      man_cal: el(c, "man_cal").checked,
+      move_time_ms: parseInt(el(c, "move_time_ms").value),
+      move_time_limit_pos_ms: parseInt(el(c, "move_time_limit_pos_ms").value),
       swap_inputs: el(c, "swap_inputs").checked,
       swap_outputs: el(c, "swap_outputs").checked,
     };
@@ -497,9 +500,20 @@ function updateComponent(cd) {
       setValueIfNotModified(el(c, "name"), cd.name);
       el(c, "state").innerText = cd.state_str;
       selectIfNotModified(el(c, "in_mode"), cd.in_mode);
+      checkIfNotModified(el(c, "man_cal"), cd.man_cal);
       checkIfNotModified(el(c, "swap_inputs"), cd.swap_inputs);
       checkIfNotModified(el(c, "swap_outputs"), cd.swap_outputs);
-      if (cd.cal_done == 1) {
+      if(el(c, "man_cal").checked) {
+        el(c, "man_cal_warning").style.display = "inline";
+        el(c, "man_cal_settings").style.display = "block";
+      }
+      else {
+        el(c, "man_cal_warning").style.display = "none";
+        el(c, "man_cal_settings").style.display = "none";
+      }
+      setValueIfNotModified(el(c, "move_time_ms"), cd.move_time_ms);
+      setValueIfNotModified(el(c, "move_time_limit_pos_ms"), cd.move_time_limit_pos_ms);
+      if (cd.cal_done == 1 || cd.man_cal == 1) {
         if (cd.cur_pos != cd.tgt_pos) {
           el(c, "pos").innerText = cd.cur_pos + " -> " + cd.tgt_pos;
         } else {
--- a/mos.yml
+++ b/mos.yml
@@ -66,9 +66,11 @@ config_schema:
   - ["wc.swap_inputs", "b", false, {title: "Swap inputs (2 - open, 1 - close)"}]
   - ["wc.swap_outputs", "b", false, {title: "Swap outputs (2 - open, 1 - close)"}]
   - ["wc.calibrated", "b", false, {title: "Calibration done"}]
+  - ["wc.man_cal", "b", false, {title: "Manual Calibration active"}]
   - ["wc.idle_power_thr", "d", 10.0, {title: "Power consumption threshold for motor idle detection"}]
   - ["wc.move_power", "d", 0.0, {title: "Power consumption during movement, watts"}]
   - ["wc.move_time_ms", "i", 0, {title: "Move time, in millseconds"}]
+  - ["wc.move_time_limit_pos_ms", "i", 0, {title: "Move time to limit position, in milliseconds"}]
   - ["wc.current_pos", "d", 0.0, {title: "Last position, percent: 0 - fully closed, 100 - fully open."}]
 
   - ["gdo", "o", {title: "Garage door settings", abstract: true}]
--- a/src/shelly_hap_window_covering.cpp
+++ b/src/shelly_hap_window_covering.cpp
@@ -37,7 +37,8 @@ WindowCovering::WindowCovering(int id, Input *in0, Input *in1, Output *out0,
       state_timer_(std::bind(&WindowCovering::RunOnce, this)),
       cur_pos_(cfg_->current_pos),
       tgt_pos_(cfg_->current_pos),
-      move_ms_per_pct_(cfg_->move_time_ms / 100.0) {
+      move_ms_per_pct_(cfg_->move_time_ms / 100.0),
+      move_limit_ms_per_pct_(cfg_->move_time_limit_pos_ms / 100.0) {
   if (!cfg_->swap_inputs) {
     in_open_ = in0;
     in_close_ = in1;
@@ -168,7 +169,12 @@ Status WindowCovering::Init() {
   if (cfg_->calibrated) {
     LOG(LL_INFO, ("WC %d: mp %.2f, mt_ms %d, cur_pos %.2f", id(),
                   cfg_->move_power, cfg_->move_time_ms, cur_pos_));
-  } else {
+  } 
+  else if(cfg_->man_cal) {
+    LOG(LL_INFO, ("WC %d: mp %.2f, mt_ms %d, mtlpos_ms %d, cur_pos %.2f", id(),
+      cfg_->move_power, cfg_->move_time_ms, cfg_->move_time_limit_pos_ms, cur_pos_));
+  }
+  else {
     LOG(LL_INFO, ("WC %d: not calibrated", id()));
   }
   state_timer_.Reset(100, MGOS_TIMER_REPEAT);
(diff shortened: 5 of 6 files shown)


timoschilling:
I think it’s not a good idea to depend on a external library just for an icon.

juangamnik:
I think that the whole "Caution" is unnecessary since the calibration expects, that the motor safely finds its limit position, too. So no need for a Caution. From my point of view this can be removed without replacement :)

vaenror:
Any news on this? Awaiting this feature eagerly!
Thanks for all the hard work!

iBlackyUE:
Me too, I wait impatiently, because the automatic calibration only gives me 20% of the Shutter.
This homekit firmware is awesome.

philmue1988:
hi @rojer are there any news regarding the implementation? would be great!

rojer:
no update yet

philmue1988:
@rojer thanks for your effort! Could you estimate, when it will be possible? I have the same problem that I cant use the shelly like I want to due to the fact that the consumption is 1.083 and stop after 1 sec instead of 28.

rojer:
sorry, no ETA. i'm focusing on other things at the moment.

philmue1988:
okay thank you. I haven't found a walkaround so far. Do you know if there would be one?

rojer:
you could always use a custom build, i believe @juangamnik posted a link to firmware with this patch before.

juangamnik:
> @juangamnik entweder bin ich zu blöd... siehe Screenshot.
> 
> Hab's dann selber per Fork probiert, jedoch scheitert der Buildprocess.. könntest du mir deine Firmware irgendwie anders zukommen lassen? :)
> 
> CleanShot 2021-10-29 at 15 24 47 .

Dear @raaaf. The artifacts are removed after some time. I couldn't rebuild the project as is either (although this is done in docker and should be repeatable...), but I synced with the upstream repository and did a rebuild and that completed. But be careful... I did not test the new build yet!

Here you can find the release:
https://github.com/juangamnik/shelly-homekit/actions/runs/1403210981

raaaf:
@juangamnik cheers, mate. It works perfectly and finally I have the correct visual ui in Apple Home. Thanks for writing the necessary code and rebuilding it. I appreciate it.

image

juangamnik:
@raaaf kewl… thanks you‘re welcome

vaenror:
Thank you very much for updating 🙏
I've only used the OTA method to install/update my shellys. How do I install this fork, dies it work OTA? I tried uploading the.zip from this fork/zipball/master but that doesn't work.

Thanks in advance.

rojer:
thanks @juangamnik !
i uploaded the shelly 2.5 zip to my site, you can use this url for OTA: http://rojer.me/files/shelly/misc/shelly25-449.zip

vaenror:
Thank you guys very much 🙏

Kerubin89:
Hello 
thanks for your firmware. 
I tested your Firmware to set the max. close position for my shutter. I have a catdoor in my window glass. My target is when I close the shutter with the physic button and the My home button on IPhone then is the max close position on ca.80% (perfect postion that can cat go out and in with a close shutter). With your firmware works for one time to close. But when I open the shutter, then is not full open and when I close agan, the position is not the same then before. 
I think the problem is, the close time and the open time is not the same. Can you make a option for buth? Or I can set this on a diffrent way?

Thank you for your efforts.

vaenror:
Tried the build but somehow it's not working for me. My Shutters need around 22s to close. And 23 to open. But even with 22000ms the shutters are not opening or closing correctly and in homekit it's always showing the opposite state. Weird.

flomyro:
> 
> Dear @raaaf. The artifacts are removed after some time. I couldn't rebuild the project as is either (although this is done in docker and should be repeatable...), but I synced with the upstream repository and did a rebuild and that completed. But be careful... I did not test the new build yet!
> 
> Here you can find the release: https://github.com/juangamnik/shelly-homekit/actions/runs/1403210981

Thanks so much @juangamnik!! I was stuck on Version 2.7.3 and the timeout Bug after 28 days was driving me nuts...

juangamnik:
[Here a new release](https://github.com/juangamnik/shelly-homekit/actions/runs/1409941496)
[Here you can find a description what has changed](https://github.com/mongoose-os-apps/shelly-homekit/issues/404#issuecomment-956799061)

lucaspinelli85:
I have tested last version and I have the following problems:
1: I have an effective closing time of 33 seconds and an effective opening time of 27 seconds. I workaround by setting longest time and more, but after some up and down movements, the position is out of phase and cannot physically return completely closed or open.
2: in the opening, in the webui it does not reach 100%, but at 99%, in the homekit it remains "I'm opening ..."imageimage
3:

juangamnik:
@lucaspinelli85 @Kerubin89 @vaenror your issue seems to be about differing open and closing times and not about manual calibration because the roller shutter switches do not directly put current into the shutter motor. This was the initial issue and for that, this fork has a fix.

I believe your issues should be fixed in the automatic calibration first (introducing open and closing times, which is currently not available). Afterwards we can enhance the manual calibration to support this too. But this are two different issues, which should be handled separately. If we try to change multiple things at one time this can cause other issues (in Software Engineering small change sets are a way to handle complexity and build reliable software).

Hence, as far as I am concerned, we should merge this pull request for its original purpose as soon as @rojer says it is time and handle your problem in separate issues. OK?

lucaspinelli85:
I'm trying to make it clear that it will be impossible to have a manual calibration with a single time because even if only slightly, there will always be a small difference between there and back.
Imagine having a Ferrari on a highway that goes one way, to become a Fiat again. With the same time, Ferrari will go further so if you start from one point with the Ferrari and return with the Fiat using the same time you will never return to the starting point, multiply this fact more times and you will find yourself further and further away from the starting point.
I hope I explained myself...

juangamnik:
> I'm trying to make it clear that it will be impossible to have a manual calibration with a single time because even if only slightly, there will always be a small difference between there and back.
> Imagine having a Ferrari on a highway that goes one way, to become a Fiat again. With the same time, Ferrari will go further so if you start from one point with the Ferrari and return with the Fiat using the same time you will never return to the starting point, multiply this fact more times and you will find yourself further and further away from the starting point.
> I hope I explained myself...

Sorry that I repeat myself: The automatic calibration does not have two times, too. This are two different issues. The issue that you describe are even a problem if you have no difference between open and close time, because motors do need time to start and stop. The calculation of ramp up and down-times is not exact. Therefore the position will change over time anyway. The automatic calibration deals with this via a separate "run to limit" variant if you do not set the shutters to a value between ]0..100[, but exactly to 0 or 100. Then it will stop as soon as no current is measured. 

Since this is not possible in manual calibration (as there is no current to measure), we have the limit time. So if you go to exactly 0 or 100 it will go for the time from current position to the end position plus the limit time (to be sure to reach the end position and start from a clean state). This does mitigate the problem of diverging positions. So it "recalibrates" each time you move completely up or down. Some control software even does these recalibrations from time to time on its own (but not this firmware).

Did you try to use the Limit position parameter accordingly? For me it is `Movement Time: 11000ms` and `Limit Time: 1000ms`.

I hope I explained myself 😇

lucaspinelli85:
Of course, you have explained yourself perfectly.
Perhaps the problem is because automatic calibration is usually performed almost always in the rolling shutters which by their nature, at least once a day, do the entire opening stroke and the entire closing stroke.
So it is as if each time the percentages are recalibrated at each limit switch.
However, if you stay in the middle without touching the limit switches, the problem arises ...
As you say, it is a different request from the one for which the manual calibration was made, but in my opinion it is something to integrate as many will have this "problem" that in the automatic calibration we do not realize because we normally close the shutter all night and we reopen it all day.
If I put the time limit (maybe it's another bug) it adds up to the movement time, not only to get to the limit switch but also in the intermediate positions. Example: I set a time limit of 1 second, I open to 50%, when it reaches 50% it continues for another second. This shouldn't happen, should that second only be used when I fully open or fully close to be able to "press" the limit switch contact?

juangamnik:
> Of course, you have explained yourself perfectly.
> Perhaps the problem is because automatic calibration is usually performed almost always in the rolling shutters which by their nature, at least once a day, do the entire opening stroke and the entire closing stroke.
> So it is as if each time the percentages are recalibrated at each limit switch.
> However, if you stay in the middle without touching the limit switches, the problem arises ...
> As you say, it is a different request from the one for which the manual calibration was made, but in my opinion it is something to integrate as many will have this "problem" that in the automatic calibration we do not realize because we normally close the shutter all night and we reopen it all day.

For the roller shutter it makes no difference.

> If I put the time limit (maybe it's another bug) it adds up to the movement time, not only to get to the limit switch but also in the intermediate positions. Example: I set a time limit of 1 second, I open to 50%, when it reaches 50% it continues for another second. This shouldn't happen, should that second only be used when I fully open or fully close to be able to "press" the limit switch contact?

Are you sure, [that you use this release](https://github.com/juangamnik/shelly-homekit/suites/4225829277/artifacts/109336061)? This was fixed in this version as mentioned above

lucaspinelli85:
But I am not using the shelly in manual calibration for a roller shutter, but for the motorization of an elevator that I cannot connect directly to the motor, but via the control unit! For this reason I have these problems, because I don't often go to the limit switches and stay in the middle with the movements.
I use the latest version @rojer  shared yesterday.

lucaspinelli85:
I think I understand problem 3.
If I raise / lower the homekit button it works, but if I raise / lower the percentage bar or even directly from webui, the control unit goes into alarm as if I had operated up and down at the same time.
In my opinion this happens because the exchange of the output from up to down and vice versa is so fast and sudden that the control unit perceives it as a signal at the same time and therefore goes into alarm.
This does not happen if I use the button in the homekit because there is a time lag between one action and another (there is a stop, instead of going from up directly to down and vice versa).
It would take a few milliseconds of interval between the up and down action and vice versa, or interpose the stop between one and the other so as not to pass directly from one condition to another.

juangamnik:
> I think I understand problem 3.
> If I raise / lower the homekit button it works, but if I raise / lower the percentage bar or even directly from webui, the control unit goes into alarm as if I had operated up and down at the same time.
> In my opinion this happens because the exchange of the output from up to down and vice versa is so fast and sudden that the control unit perceives it as a signal at the same time and therefore goes into alarm.
> This does not happen if I use the button in the homekit because there is a time lag between one action and another (there is a stop, instead of going from up directly to down and vice versa).
> It would take a few milliseconds of interval between the up and down action and vice versa, or interpose the stop between one and the other so as not to pass directly from one condition to another.

The firmware is implemented as a state machine which has besides others different states for stopped, ramp-up, moving, and ramp-down. It might be, that the settings for ramp-up and ramp-down do not match your setup. @rojer is the ramp-up/-down depending on the flowing current? Perhaps this abrupt changes are because the state change is two fast without a value for current?

raaaf:
@juangamnik @rojer we just moved to a new flat and I need the modified firmware again.. sadly, I can't find it on your server (@rojer). 

Can you provide it again (@rojer) or can you build the artifact and provide it again, @juangamnik? :)

Cheers guys!

raaaf:
@juangamnik @rojer Help a fella out, please ;)

juangamnik:
> @juangamnik @rojer Help a fella out, please ;)

@raaaf sorry, but the file has been erased from GitHub automatically IIRC. @rojer had a link to the firmware AFAIK. Can you send it again?

Or else I need to do a new build but I am unsure when I can do this 😰

raaaf:
How about right now? :)
Just kidding – though having the possibility to calibrate my shutters again is much appreciated.

Cheers!

rojer:
@raaaf built and uploaded to https://rojer.me/files/shelly/misc/shelly25-449.zip

raaaf:
@rojer awesome – thanks for your effort.. however the manual calibration part is missing or did it change? :(
CleanShot 2022-05-18 at 09 43 03

rojer:
@raaaf it definitely is there, i just flashed and verified. check the build id under system settings, make sure it's `20220517-212055/2.10.1-10-gae6dacd-p449`

raaaf:
@rojer damn it.. accidentally flashed the latest beta (browser url cache).. cheers and keep it up!

PS: if you need any support in terms of design, give me a ping. I'd love to give back!

greensouth:
One lil question... does this work for rolling shutters with BIG rollers, like spanish ones? That suppose that time for opening and closing is not linear, because with every turn the rollers gets bigger exponentially. That makes the position not accurate at all. I think that there is another issue covering that subject.

simonbrueckner:
First of all, thanks a million for the HomeKit firmware in general and this addition to tweak calibration.
I tried the firmware, however I am also running into the stopping at 99% open issue when using the manual calibration.
I believe that several people (including myself) have a different Problem as is tackled here:
For me at least, the issue is that my roller motor stops correctly at the top position, but at the bottom position, it will still roll even if the blinds are already full closed, leading to an ugly rattle noise until the motor stops. My blinds are fully closed at around 15%.
So I set a scene stopping closing the blinds at 15%. Doing so, I am not seeing the calibration going off, probably because the stock calibration only is used to indicate the rough position and will actually realign each time the motor reaches one end position.
So for me, I’d actually need two input fields to limit top or bottom positions to a given percentage value, and letting the other one go to the final position. Doing so, the calibration would stay intact and I could still use the 0-100% in HomeKit.

lb-supertec:
I have multiple window openers GEZE ECChain which i want to control with a shelly 2.5.
I never managed to calibrate them in roller shutter mode.
The reason might be the low energy consumption of the motors of only 2 watts.

I tried many things (RC snubbers, Simulating higher energy consumption with bulbs and switches).
What in the end worked best for me is the firmware shelly25-449 with manual calibration.
I would be more then happy if this would be added to the regular firmware.

Nately3:
Does the roller-shutter-mode work when the Shelly_2.5 is running at 24 Volt?

lb-supertec:
hi

i have never tried at 24V.
i also don’t have 24V nearby where the shelly is installed.
do you think this might help?

best regards
daniel

Nately3:
Hi,

my question did not refer to your problem. I need to control a 24V motor in a 24V enviroment. So I would need roller-shutter-mode with manual calibration at 24V. I am not sure if this is possible at all.

ingski:
Great Job. This alpha version solved my issue :-) Thanks a lot

Ingo

dev2ek:
First of all, I would like to thank you for this great project 😊👍👍👍😊

@rojer are there any thoughts from you as to when this great feature can be integrated into the current firmware?

I am using the current version from your link:

http://rojer.me/files/shelly/misc/shelly25-449.zip

Unfortunately this version still has the 99% problem ☹

Thank you very much for your work!

GEORGzer:
I just registered because I wanted to say that this custom firmware actually solved a problem I had with the calibration. The ability to set a custom time for that spared me a ton of headaches 👍

I think it should be a part of the official release to give people the option.

juangamnik:
> I just registered because I wanted to say that this custom firmware actually solved a problem I had with the calibration. The ability to set a custom time for that spared me a ton of headaches 👍
> 
> I think it should be a part of the official release to give people the option.

Thanks

GEORGzer:
> 

You‘re welcome!

Unfortunately I‘ve run into the „99%“ problem as well, where steh shelly stops at „idle 99“ and the status in Homekit keeps spinning at „opening“ 🤷‍♂️

GEORGzer:
> > I just registered because I wanted to say that this custom firmware actually solved a problem I had with the calibration. The ability to set a custom time for that spared me a ton of headaches 👍
> > I think it should be a part of the official release to give people the option.
> 
> Thanks

You‘re welcome!

Unfortunately I‘ve run into the „99%“ problem as well, where the Shelly stops at „idle 99“ and the status in Homekit keeps spinning at „opening“ 🤷‍♂️

juangamnik:
> Unfortunately I‘ve run into the „99%“ problem as well, where the Shelly stops at „idle 99“ and the status in Homekit keeps spinning at „opening“ 🤷‍♂️

Yeah, I have this issue to. I am not sure what is causing this. Perhaps I misunderstood the state machine „behind“ the firmware and missed a case. If someone finds my mistake, feel free to correct me.

desek90:
I really hope there will be soon some progress on the 99% issue, because that feature is exactly what I need for my roller blinds
Keep up that excellent work guys!

nicx:
Is there any progress to this? I Would love to see this, so I do not need to buy new switches like the Nous B3T supporting matter.

Issue #379, opened 2021-01-03
Publish automatic latest release


rojer:
Add a GitHub action to automatically build a release from master
Github release should be called 'latest'
Version should be incremented with number of commits from latest tag.
This can be taken from `git describe output`:
For example, right now we are 7 commits past 2.7.0:

```
$ git describe --tags
2.7.0-7-g3cfc021
```

Current latest release should have version `2.7.0-latest7`

This will ensure version check comparison during automatic update check works.

timoschilling:
The GitHub release should be called a prerelease if it's not a stable one.

ZeevoX:
I have started work on a pull request but have not implemented the part of the GitHub workflow to upload the produced firmware files, as I would like to ask for clarification on how this ought to be done. 

If we have a 'latest' GitHub release that would imply tagging every commit on master so that it always pointed to the latest commit. Should the script instead attach files to the last tag? (e.g. attaching `2.7.0-latest7` to the `2.7.0` release) Or could the script instead upload the firmwares to `rojer.me` like @rojer manually did [here](https://github.com/mongoose-os-apps/shelly-homekit/issues/50#issuecomment-753633174)?

timoschilling:
Maybe we can remove rojer.me and direct work with the GitHub release api?

andyblac:
thats how it was originally, but was moved to Roger.me for better controll.

timoschilling:
What was the concrete reason / problem?

andyblac:
not sure was rojer decision.

andyblac:
think it was so he could control beta releases.

rojer:
we need a plain HTTP location for stock -> HK update as stock firmware can't flash from HTTPS location (at least not from github).
we could use github urls for HK -> HK updates but for me it's just easier to have everything in one place.

andyblac:
it also makes it easier to maintain script.

PR #450, opened 2021-01-29
Add GitHub action

ZeevoX:
* Add a GitHub action that automatically builds firmware for supported Shelly devices (#379)
* Uses `git describe` to put firmware zips into an appropriately-titled directory e.g. `2.7.3-latest8-geee7898` 

View a sample build [here](https://github.com/ZeevoX/shelly-homekit/runs/1788742790).

### DIFF
--- a/.github/version.py
+++ b/.github/version.py
@@ -1,0 +1,48 @@
+#!/usr/bin/python3
+import argparse
+import subprocess
+
+def run(*args):
+    return subprocess.check_output(args).decode('utf-8').strip()
+
+def git_describe(*args):
+    return run('git', 'describe', '--tags', *args)
+
+def git_branch_name():
+    return run('git', 'branch', '--show-current')
+
+def commits_since_forkpoint():
+    return int(run('git', 'rev-list', '--count',  '--no-merges', 'origin/master..'))
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Get Shelly-HomeKit version name using Git')
+    parser.add_argument('-s', '--suffix', action='store_true', help='output only the suffix after the version number')
+    parser.add_argument('-r', '--release', action='store_true', help='just use tag number, for stable releases')
+    args, extra_args = parser.parse_known_args()
+
+    # all unknown command line arguments are passed directly to git-describe
+    tag_name, commits_since_tag, extra = git_describe(*extra_args).split(sep='-', maxsplit=2)
+
+    # usually this is just the sha of the git commit
+    # however if e.g. --dirty is passed then, if working tree
+    # dirty, it will be contained in the `info` list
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -8,7 +8,7 @@ LOCAL ?= $(shell which docker> /dev/null && echo -n 1 || echo -n 0)
 CLEAN ?= 0
 VERBOSE ?= 0
 RELEASE ?= 0
-RELEASE_SUFFIX ?=
+RELEASE_SUFFIX ?= $(shell git describe --tags --dirty | sed -r 's/^[^-]*(-|$$)/-latest/g')
```
rojer, line 11: leave it empty by default, set from the action as needed

.github/workflows/build_shelly_homekit.yaml
```diff
@@ -0,0 +1,25 @@
+on:
+  push:
+  pull_request:
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
rojer, line 6: we have docker containers for mos - repository name is `mgos/mos`, could we use that?

ZeevoX: I've given it a go, however, I've run into a problem: a local build requires docker – which, from what I gather, would require running docker in a docker container. In this context, what improvements does using docker image hold over ubuntu with mos ppa?

timoschilling: I agree with @ZeevoX, his current way is the better one.

.github/version.py
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/python3
+import argparse
+from subprocess import check_output
```
rojer, line 3: don't use "from" imports, import the module itself and refer to its contents, i.e. `subprocess.check_output`

rojer (commented):
LGTM

drop index.html.gz, i think it was picked up by mistake

rojer (approved):

ZeevoX:
I've added the uploading of built firmware zips to the GitHub action. They can be downloaded from the _Artifacts_ section on the job summary page. You can view a sample build [here](https://github.com/ZeevoX/shelly-homekit/actions/runs/541941571).

timoschilling:
I think it would be cool to provide them as separate downloads and not as one

ZeevoX:
Unfortunately this a known limitation of GitHub Actions, see [here](https://github.com/actions/upload-artifact#zipped-artifact-downloads). Although we could upload separately-named artifacts for each firmware, not only would we have to keep an up-to-date list of Shelly devices that we support in the action, but also each firmware zip would still be located inside another zip file.

rojer:
ok, let's give this a go and see how it works in practice.

PR #458, opened 2021-02-02
fix hap code validation

timoschilling:
when `hapSetupCode` is for example `"11223344"`
then `codeMatch` is `["11223344", "112", "23", "344"]`
the current `slice(1, 3)` returns `["112", "23"]`
`slice(1, 4)` or the sort version `slice(1)` returns `["112", "23", "344"]`.


rojer (approved):
LG but please update .gz as well

timoschilling:
Will do after the "ui-confirm" branch is merged

PR #459, opened 2021-02-03
remove unnecessary state read

timoschilling:
The result is unused and `GetState` has no side effects, so I assume that is unnecessary.


andyblac:
is looks like it is used here https://github.com/mongoose-os-apps/shelly-homekit/blob/070468d17d678654aecd9525ac9ff34942f9cbb9/src/shelly_switch.cpp#L226

timoschilling:
🤦‍♂️

PR #462, opened 2021-02-04
add invert output option

### LINKED ISSUE #408: Add: Inverted Output
can we add an option to the "Switch" mode for "Inverted Relay" (maybe only for the Shelly1, as that has a Dry Contact Relay and can be used for various things other than a power relay)

Description:
So the HomeKit icon shows as ON when the relay is OPEN (OFF), and HomeKit icon shows as OFF when relay is CLOSED (ON).

as per disscussion here https://github.com/mongoose-os-apps/shelly-homekit/issues/406

timoschilling:
closes #408

### DIFF
--- a/mos.yml
+++ b/mos.yml
@@ -41,6 +41,7 @@ config_schema:
   - ["sw", "o", {title: "Switch settings", abstract: true}]
   - ["sw.name", "s", "", {title: "Name of the switch"}]
   - ["sw.enable", "b", true, {title: "Enable this switch in the accessory"}]
+  - ["sw.out_inverted", "b", true, {title: "Invert output, set to true for normally open output"}]
   - ["sw.in_mode", "i", 1, {title: "-1 - Absent, 0 - Momentary, 1 - Toggle, 2 - Edge, 3 - Detached"}]
   - ["sw.in_inverted", "b", false, {title: "Invert input, set to true for normally closed input"}]
   - ["sw.state", "b", false, {title: "State of the switch"}]
--- a/fs_src/index.html
+++ b/fs_src/index.html
@@ -304,6 +304,13 @@
         </select>
       </div>
       <div class="form-control">
+        <label for="out_inverted">Inverted output:</label>
+        <label class="switch">
+          <input type="checkbox" id="out_inverted">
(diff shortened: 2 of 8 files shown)


andyblac:
another one need a conflict correction 😄

mos.yml
```diff
@@ -40,6 +40,7 @@ config_schema:
   - ["sw", "o", {title: "Switch settings", abstract: true}]
   - ["sw.name", "s", "", {title: "Name of the switch"}]
   - ["sw.enable", "b", true, {title: "Enable this switch in the accessory"}]
+  - ["sw.inverted", "b", true, {title: "Invert output, set to true for normally open output"}]
```
rojer, line 43: we have `in_inverted`, let's call this option `out_inverted`

rojer (commented):
overall looks good but there is one more thing - when constructing components, we need to reset output to non-inverted as a common baseline, so for example if you change switch into window covering, that know nothing about inverting outputs, outputs do not remain inverted.
it's not ideal, and represents a leak in abstractions, so if you have better idea how to manage it - welcome.
but for now, we'll need something similar to [this](https://github.com/mongoose-os-apps/shelly-homekit/blob/c335332eb96c05d84411d9a67d7929cabafd620c/src/shelly_main.cpp#L318-L320).

timoschilling:
Maybe I miss something, but I think we don't need to reset it. `inverted` is a config property of `sw`, which means that if the user switches from `sw` to `window` it reboots the devise which means a new output is created and which will use the `wc` setting and will never be in `inverted` mode.

rojer:
> it reboots the devise

except it doesn't. `*restart_required = true` does not mean reboot, it means HAP server restart.
HAP server gets stopped, components get destroyed and re-created. however, peripherals - the Inputs, Outputs, temp sensors, etc, do not. so if inverted setting would be preserved and new components created with inverted output(s).

timoschilling:
review applyed

rojer:
i don't see changes to shelly_main

timoschilling:
thanks for the hint, lost during a rebase

rojer:
merged, thanks!

PR #470, opened 2021-02-07
use BTN_NOISY to decide useing NoisyInputPin or InputPin

timoschilling:
At the moment `BTN_GPIO` will configure `SetupButton` which uses `NoisyInputPin` on ESP8266. Thats a problem for the upcoming RGBW2 support. RGBW2 needs PWM and PWM uses the hardware timer, but the hardware timer is blocked by `NoisyInputPin`. 

In a gitter chat @rojer says that only 1L and 1PM needs `NoisyInputPin`, all other can use `InputPin`.

This PR explizit configures the use of `NoisyInputPin` or `InputPin` for `SetupButton`. That allow the `SetupButton` use and PWM on the RGBW2 at the same time.

### DIFF
--- a/mos.yml
+++ b/mos.yml
@@ -99,6 +99,7 @@ cdefs:
   PRODUCT_MODEL: '"${build_vars.MODEL}"'
   LED_GPIO: -1
   BTN_GPIO: -1
+  BTN_NOISY: 0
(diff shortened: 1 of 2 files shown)


rojer:
makes sense, thanks!

PR #472, opened 2021-02-07
load donate button only once

timoschilling:
Before this change the PayPal donate button was loaded after each web socket response / each sec. With this change it is only loaded once.

Screenshot 2021-02-07 at 17 47 50


rojer:
good catch, thanks!

rojer (approved):

andyblac:
@timoschilling pesky index again :(

rojer:
thank you!

PR #474, opened 2021-02-08
[Flashscript] version 2.2.2

andyblac:
[FlashScript] add --type, this allow scan to limit it to firmware type (stock, homekit or all)
[FlashScript] add --ip-type, allows config of Static IP or DHCP.
[FlashScript] move hap-setup-code out of write_flash.
[FlashScript] update usage section.

### DIFF
--- a/tools/flash-shelly.py
+++ b/tools/flash-shelly.py
@@ -691,7 +691,7 @@ def device_scan(hosts, action, dry_run, quiet_run, silent_run, mode, type, exclu
       if deviceinfo.fw_type is not None and (deviceinfo.fw_type in type or type == 'all'):
         device = {'host': deviceinfo.host, 'wifi_ip': deviceinfo.wifi_ip, 'fw_type': deviceinfo.fw_type, 'device_url': deviceinfo.device_url, 'info' : deviceinfo.info}
         total_devices += 1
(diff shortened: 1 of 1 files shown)


tools/flash-shelly.py
```diff
@@ -372,6 +381,47 @@ def is_newer(v1, v2):
...
 
+def write_static_ip(device_info, network_type, ipv4_ip='', ipv4_mask='', ipv4_gw='', ipv4_dns=''):
+  wifi_ip = device_info.wifi_ip
+  if device_info.fw_type == 'homekit':
+    if network_type == 'static':
+      message = f"Configuring static IP to {ipv4_ip}..."
+      mask_value={'config': {'wifi': {'sta': {'netmask': ipv4_mask}}}}
+      dns_value={'config': {'wifi': {'sta': {'nameserver': ipv4_dns}}}}
+      gw_value={'config': {'wifi': {'sta': {'gw': ipv4_gw}}}}
+      ip_value={'config': {'wifi': {'sta': {'ip': ipv4_ip}}}}
```
timoschilling, line 392: Why not this way?
```suggestion
      value={'config': {'wifi': {'sta': {'netmask': ipv4_mask, 'nameserver': ipv4_dns, 'gw': ipv4_gw, 'ip': ipv4_ip}}}}
```
and then one single request to save it?

andyblac: yeah, that is a lot cleaner.

PR #476, opened 2021-02-09
improve socket logging

timoschilling:
Before:
We have a dedicated logging on some points and a explizit logging if the message send sucessed.

After:
We have always a logging before the message sends and a sucessed / error message after.

### DIFF
--- a/fs_src/script.js
+++ b/fs_src/script.js
@@ -38,7 +38,6 @@ el("sys_save_btn").onclick = function () {
       sys_mode: parseInt(el("sys_mode").value),
     },
   };
-  console.log("sysSetCfg:", data);
   el("sys_save_spinner").className = "spin";
   pauseAutoRefresh = true;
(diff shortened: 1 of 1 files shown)


fs_src/script.js
```diff
@@ -815,16 +812,17 @@ function sendMessageWebSocket(method, params = [], id = requestID) {
   requestID += 1
   return new Promise(function (resolve, reject) {
     try {
+      console.log("[send] " + method + " " + id + ":", params);
       socket.send(JSON.stringify({"method": method, "id": id, "params": params}));
-      console.log("[send] Data sent: " + JSON.stringify({"method": method, "id": id, "params": params}));
+      console.log("[sent] " + method + " " + id);
```
andyblac, line 817: missing `params` ?

timoschilling: as this is only the done message with the same id as the log from 2 lines above I don't want to repeat it

andyblac: ok

Issue #475, opened 2021-02-09
What are the browser requirements of this project?


timoschilling:
Does we still support IE11 for example?

rojer:
um... no. no, no, no. :)

since it's targeted at Apple users, obviously safari support, mobile and desktop, is a must.
chrome also should work, the rest - don't care, pretty much.

timoschilling:
Ok nice

PR #477, opened 2021-02-10
use js string templates

timoschilling:
This PR switches from js string concatenation to string templates aka string interpolation.
As Rojer says in #475 that we don't want to support IE 11 this works in any current browser.


rojer:
nice, thanks!

PR #480, opened 2021-02-11
Css cleanup

timoschilling:
Remove some css 💩, saves us some bytes.

We don't need the removed vendor prefixes (`-webkit-`/`-ms`) on our supported browsers.

This changes the layout a bit on non apple devises as it's no longer uses the Light variants of the font, but on apple devises they have never been used.

I didn't find any optical change in this PR on latest Mac Safari


rojer:
thanks for the cleanup!

PR #483, opened 2021-02-12
improve button css

timoschilling:


fs_src/style.css
```diff
@@ -25,6 +25,12 @@ button {
   border-radius: 18px;
   border: none;
   cursor: pointer;
+  background: #0071e3;
+}
+
+button label {
+  position: relative;
+  cursor: pointer;
```
timoschilling, line 33: Without this `cursor: pointer;` the mouse pointer was a hand on the button but a arrow on the button label

fs_src/index.html
```diff
@@ -43,7 +43,7 @@ <h1 class="">General Settings</h1>
       </div>
       <div class="form-control">
         <label></label>
-        <button class="btn" id="sys_save_btn">
```
timoschilling, line 46: we do element styling now, so no need for class attachments

rojer (approved):
LGTM, please rebase

timoschilling:
rebased

PR #484, opened 2021-02-12
Js cleanup

timoschilling:


fs_src/script.js
```diff
@@ -1,19 +1,3 @@
-var lastInfo = null;
-
-var socket = null;
-
-var wifiEn = el("wifi_en");
```
timoschilling, line 5: We use global element variables only for wifi and for hap, why? I think we should do it no where, use the same coding style every where.

rojer: legacy. you are right, should just use `el()`

rojer (approved):

PR #496, opened 2021-02-13
[webui] allow auto refresh to see if update has happened

### LINKED ISSUE #487: update badge does not auto update correct status
the update badge does not auto show hide on auto refresh
Ie

1. when on 2.7.2
2. open webUI
3. used tools/flash-shelly.py to update, but you can confirm by have 2 webui windows open, one to update, one to show status bug.
4. after device reboot
5. webui still shows “update”
6. refresh page
7. update badge gone

andyblac:
fixes #487

### DIFF
--- a/fs_src/script.js
+++ b/fs_src/script.js
@@ -849,9 +849,14 @@ function onLoad() {
       // check for update only once when loading the page (not each time in getInfo)
       if (lastInfo.wifi_rssi !== 0) {
         var last_update_check = parseInt(getCookie("last_update_check"));
+        var cached_curVersion = getCookie("cached_curVersion");
+        var curVersion = lastInfo.version
         var now = new Date();
         console.log("Last update check:", last_update_check, new Date(last_update_check));
         if (isNaN(last_update_check) || now.getTime() - last_update_check > 24 * 60 * 60 * 1000) {
+          checkUpdate();
+        }
+        if (curVersion > cached_curVersion) {
           checkUpdate();
         }
         el("notify_update").style.display = (getCookie("update_available") ? "inline" : "none");
@@ -1016,6 +1021,7 @@ async function checkUpdate() {
         if (cfg.urls) updateURL = cfg.urls[model];
       }
       console.log("Version:", latestVersion, "URL:", updateURL);
+      setCookie("cached_curVersion", curVersion);
       if (!latestVersion || !updateURL) {
         console.log("Update section not found:", model, curVersion, cfg);
         e.innerHTML = errMsg;


fs_src/script.js
```diff
@@ -849,11 +849,16 @@ function onLoad() {
       // check for update only once when loading the page (not each time in getInfo)
       if (lastInfo.wifi_rssi !== 0) {
         var last_update_check = parseInt(getCookie("last_update_check"));
+        var cached_curVersion = getCookie("cached_curVersion");
```
rojer, line 852: you don't need to use a cookie to store this, a global variable would be sufficient.

andyblac: sorry but do i do that ?

andyblac: i tried

```
var cached_curVersion = null;
async function checkUpdate() {
  var model = lastInfo.model;
.....
      }
      console.log("Version:", latestVersion, "URL:", updateURL);
      var cached_curVersion = curVersion
      if (!latestVersion || !updateURL) {
        console.log("Update section not found:", model, curVersion, cfg);
```

but after a fresh and auto refresh starts i get this.
Screenshot 2021-02-14 at 00 18 56

and update banner remains

andyblac:
tbh i dont see how we can use a global variable as the device is getting rebooted after the flash, does that not wipe the global variable ?

rojer:
you only need it to decide when to reload the page.
create `var initialVersion;` on the global level
after successful GetInfo, if it is not set yet (=== undefined), set it to data.version.
if it is set, compare to data.version and if it doesn't match, 1) reset last_update_check cookie 2) reload the page

andyblac:
> you only need it to decide when to reload the page.
> create `var initialVersion;` on the global level
> after successful GetInfo, if it is not set yet (=== undefined), set it to data.version.
> if it is set, compare to data.version and if it doesn't match, 1) reset last_update_check cookie 2) reload the page

but does it nit loose the data stored on reboot ?

rojer:
it only resets on page reload.
if firmware was updated in the background (e.g. via script), UI will lose websocket connection and then re-establish it.

andyblac:
> it only resets on page reload.
> if firmware was updated in the background (e.g. via script), UI will lose websocket connection and then re-establish it.

ok I'll try

andyblac:
hmm this what have in and get
```
        console.log("curVersion:", curVersion, "initialVersion", initialVersion);
        console.log("initialVersion0:", initialVersion);
        if (initialVersion === undefined) {
          console.log("111");
          initialVersion = curVersion;
        } else {
          console.log("222");
          if (curVersion > initialVersion) {
            console.log("333");
            checkUpdate();
          }
        }
        console.log("initialVersion1:", initialVersion);
```
Screenshot 2021-02-14 at 14 39 45

andyblac:
what section of script.js should i put it in, if not onload ?

fs_src/script.js
```diff
@@ -849,11 +849,16 @@ function onLoad() {
...
+        var curVersion = lastInfo.version
         var now = new Date();
         console.log("Last update check:", last_update_check, new Date(last_update_check));
         if (isNaN(last_update_check) || now.getTime() - last_update_check > 24 * 60 * 60 * 1000) {
           checkUpdate();
         }
+        if (curVersion > cached_curVersion) {
```
timoschilling, line 859: That didn't work, because `'1.9.0' > '1.10.0'` is `true`.

timoschilling:
Problem is already solved

PR #497, opened 2021-02-14
[Flashscript] version 2.3.1

andyblac:
hide error message "Could not resolve host:" during flash.

### DIFF
--- a/tools/flash-shelly.py
+++ b/tools/flash-shelly.py
@@ -163,7 +163,7 @@ class Device:
     self.flash_label = "Latest:"
     self.force_flash = False
 
-  def is_host_reachable(self, host):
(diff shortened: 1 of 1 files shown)


tools/flash-shelly.py
```diff
@@ -163,7 +163,7 @@ def __init__(self, host=None, wifi_ip=None, fw_type=None, device_url=None, info=
     self.flash_label = "Latest:"
     self.force_flash = False
 
-  def is_host_reachable(self, host):
+  def is_host_reachable(self, host, isflashing=False):
```
rojer, line 166: isflashing -> is_flashing, ilikemywordsdelimitedpleasethankyouverymuch :)

andyblac: okdone, is it better now 😄

PR #503, opened 2021-02-14
[Flashscript] Version 2.3.5

andyblac:
split get_remote_info separate functions for fw type
fix Host: label in listing


andyblac:
this splits remote info grab so that if eiter fails, it wont effect each others updates. only updating stock on none latest fw to homekit requires both to be online at once.

PR #509, opened 2021-02-15
fix button spinner

timoschilling:
Revert buttons to her old size and split all double button's into two rows.
image
image
(iPhone 8 size)

### DIFF
--- a/fs_src/index.html
+++ b/fs_src/index.html
@@ -77,6 +77,9 @@
         <label></label>
         <button id="hap_save_btn">
           <label><span id="hap_save_spinner"></span>Save</label>
+      </div>
+      <div class="form-control">
+        <label></label>
         </button>
         <button id="hap_reset_btn">
           <label><span id="hap_reset_spinner"></span>Reset</label>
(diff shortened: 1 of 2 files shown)


andyblac:
IMO it look better with smaller buttons, could you not move the spinner out of the button and in the top right corner of the pane or somewhere else ?

timoschilling:
@andyblac Maybe later

andyblac:
> @andyblac Maybe later

i just redone all the wiki, ok i wont redo again i'll wait to see what happens.

rojer:
can we do this only for narrow viewport sizes? we should be using horizontal space when it's available (i.e. desktop)

timoschilling:
Sorry I’m not an css expert. For me the two row button layout is only solvable with html markup changes. So I can’t use media queries to switch between. 

The options which I can build are 
1. The old version, which breaks on mobile
2. The small version with broken spinner
3. This version

maybe someone else can improve 3. In the future

rojer:
ok, let's go with this

andyblac:
there seems pleatty of space to still have them side by side, even if you increase the size, just centralise the buttons in the pane instead of having them to the right side of panel.

IMG_0508

sorry i just don't get why they are above each other.

andyblac:
and help is top pane is off 
Screenshot 2021-02-16 at 10 07 34

andyblac:
like this
107997917-41142d00-6fe4-11eb-8cca-b6001b7820e7

andyblac:
ok coded it 
Screenshot 2021-02-16 at 10 50 03
@rojer what you think, should i do PR ?

PR #523, opened 2021-02-20
remove support for valve type Shower Head and Water Faucet

### LINKED ISSUE #510: [Valve] Faucet and Shower Head broken
do not work on my Shelly1, i just get no response, as reported here https://github.com/mongoose-os-apps/shelly-homekit/issues/498#issuecomment-779412674

timoschilling:
closes #510, I can't find the problem, so I provide a fix which removes the support for valve type Shower Head and Water Faucet. Generic Valve and Irrigation will still work.

If no one can fix #510, we can remove the support of the non working types to unblock v2.8.0.

### DIFF
--- a/fs_src/index.html
+++ b/fs_src/index.html
@@ -273,8 +273,6 @@
         <select id="valve_type">
           <option id="valve_type_0" value="0">Generic Valve</option>
           <option id="valve_type_1" value="1">Irrigation</option>
-          <option id="valve_type_2" value="2">Shower Head</option>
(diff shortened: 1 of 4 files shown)


rojer:
can you leave a note somewhere as to why other types are not yet supported, linking to #510 for details?

timoschilling:
Sure, what do you think is the best place? Mos config sw.valve_type title or in the valve hap class file?

rojer:
let's put it in both places

rojer:
thanks!

PR #529, opened 2021-02-24
implement rgb support for the rgbw2

timoschilling:
This PR brings support for RGB functionality on the ShellyRGBW2.

This PR will only implement support for RGB, the W channel isn't usable with it! But will be added in a later PR, follow https://github.com/mongoose-os-apps/shelly-homekit/issues/72 for details.

Many tanks going to @andyblac for his support!

If you would like to help testing: [fw.zip](https://github.com/mongoose-os-apps/shelly-homekit/files/6185219/fw.zip)

(You can use `tools/flash-shelly.py --local-file` to flash it, you must set your RGBW2 into color mode, otherwise flashing will not work)

There are two things that are conceptually different in this PR then in the switch implementation:
1. The switch defines state as a bool, in the RGB the state is for me not only on off, for me all hap chars (on, hue, saturation and brightness) are state. Thats result some differences between switch and rgb, for example `OutputPin::SetStatePWM`.
2. Switch uses the output pin as source of truth, that's didn't work, you can't recalculate the hue and saturation from the pwm state of a pin. That's why the config is now the source of truth about the state.

I would like to add a separate `OutputPinPWM` (or what ever we name it) analog to the `OutputPin` but with a `SetState(float duty, const char *source)`. But I have some problems with that concept. I will retry it and if it's not worked I will come up with details and hope for help.


src/shelly_output.cpp
```diff
@@ -60,6 +65,17 @@ Status OutputPin::SetState(bool on, const char *source) {
   return Status::OK();
 }
 
+Status OutputPin::SetStatePWM(float duty, const char *source) {
```
timoschilling, line 68: I don't like this, but I could find any other solution without doing a big refactoring.

src/shelly_hap_rgb.cpp
```diff
@@ -0,0 +1,480 @@
...
+    // Input is active, re-arm.
+    LOG(LL_INFO, ("Input is active, re-arming auto off timer"));
+    auto_off_timer_.Reset(cfg_->auto_off_delay * 1000, 0);
+    return;
+  }
+  cfg_->state = false;
+  SetOutputState("auto_off");
+}
+
+void RGB::InputEventHandler(Input::Event ev, bool state) {
```
timoschilling, line 326: This is currently redundant to `ShellySwitch::InputEventHandler`, will move that to a separate class before this PR is finished.

rojer (commented):
looks good! only one minor comment and we can try it in alpha

mos.yml
```diff
@@ -82,6 +82,20 @@ config_schema:
   - ["gdo.move_time_ms", "i", 30000, {title: "Movement time, ms"}]
   - ["gdo.pulse_time_ms", "i", 300, {title: "Output active time, ms"}]
 
+  - ["rgb", "o", {title: "RGB settings", abstract: true}]
+  - ["rgb.name", "s", "RGB", {title: "Accessory name"}]
+  - ["rgb.enable", "b", true, {title: "Enable this switch in the accessory"}]
+  - ["rgb.in_mode", "i", 1, {title: "-1 - Absent, 0 - Momentary, 1 - Toggle, 2 - Edge, 3 - Detached"}]
+  - ["rgb.in_inverted", "b", false, {title: "Invert input, set to true for normally closed input"}]
+  - ["rgb.state", "b", false, {title: "State of the lightbulb"}]
+  - ["rgb.svc_type", 4] # Lighting
```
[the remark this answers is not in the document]
timoschilling:
```suggestion
```

timoschilling:
> we can try it in alpha

I would like to ship 2.8 before merging this, there is a little bug in it and maybe we can fix it before merging. I think 2.8 is ready to ship.

timoschilling:
@rojer can you start providing stock firmware for the rgbw2 on your server to enable stock rollback

rojer:
> @rojer can you start providing stock firmware for the rgbw2 on your server to enable stock revert

currently the [script](http://rojer.me/files/shelly/stock/update.sh) downloads everything from `https://api.shelly.cloud/files/firmware`, including SHRGBW2.zip and SHRGBWW-01.zip. what's missing?

andyblac:
> > @rojer can you start providing stock firmware for the rgbw2 on your server to enable stock revert
> 
> currently the [script](http://rojer.me/files/shelly/stock/update.sh) downloads everything from `https://api.shelly.cloud/files/firmware`, including SHRGBW2.zip and SHRGBWW-01.zip. what's missing?

for RGBW2 for stock there 2 firmwares one for colour mode, one for white, can not simply flash `SHRGBW2.zip` this why i find what the color mode is

https://github.com/mongoose-os-apps/shelly-homekit/blob/master/tools/flash-shelly.py#L373

and script gets correct firmware 
ie
`http://shelly-api-eu.shelly.cloud/firmware/SHRGBW2-color.zip` or ``http://shelly-api-eu.shelly.cloud/firmware/SHRGBW2-white.zip`
as the homekit firmware only supports color mode the stock revert url need to point the the color zip, download link above

timoschilling:
or we need to relax the update check to allow a rollback from `SHRGBW2-color` to `SHRGBW2`.

timoschilling:
depending on how we want to solve it we need to fix the UI script, it can't deal with the color/white at the moment.

timoschilling:
For the future: When we support white mode with the hk firmware too, does it make sense to rollback to color and white depending on the current configuration of the mode? I would say yes.

andyblac:
> For the future: When we support white mode with the hk firmware too, does it make sense to rollback to color and white depending on the current configuration of the mode? I would say yes.

yes, and why i originally ask you `color_mode` to be used into shellyinfo.

mrmaximas:
Is it possible to implement only one white channel (ch4) support? i'm using rgbw2 in the old white-only led fixtures with constant voltage psu for it's dimm control, connected to white channel. thanks.

andyblac:
@mrmaximas unfortunately not in this 1st stage, it is RGB only no white control at all, it will come but we need to take things one step at a time, hope you understand.

mrmaximas:
> @mrmaximas unfortunately not in this 1st stage, it is RGB only no white control at all, it will come but we need to take things one step at a time, hope you understand.

thanks. how about "roaming" feature like in latest stock firmware?

andyblac:
> > @mrmaximas unfortunately not in this 1st stage, it is RGB only no white control at all, it will come but we need to take things one step at a time, hope you understand.
> 
> thanks. how about "roaming" feature like in latest stock firmware?

create a new issue for it, and ask, thats something for rojer, as he help shelly code it for their fw.

src/shelly_main.cpp
```diff
@@ -822,6 +825,11 @@ extern "C" bool mgos_ota_merge_fs_should_copy_file(const char *old_fs_path,
       "relaydata",
       "index.html",
       "conf9_backup.json",
+      "conf3.json",
+      "hwinfo_struct.json",
```
rojer, line 829: actually, hwinfo_struct is very important. without it, stock firmware won't even run, so definitely need to keep it.
let's keep the rest of them too. i know conf3 messages are a bit annoying but i believe the file may have important factory settings that we don't want to wipe.

mos.yml
```diff
@@ -435,9 +480,13 @@ conds:
       libs:
         - origin: https://github.com/mongoose-os-libs/adc
         - origin: https://github.com/mongoose-os-libs/wifi
+        - origin: https://github.com/mongoose-os-libs/pwm
+          variant: esp8266
```
rojer, line 484: this is meaningless for a source-only library

rojer (changes requested):

src/shelly_main.cpp
timoschilling: Oh, I expected that the `hwinfo_struct` was already reverted.

timoschilling: I observed some crashes with the existing of `conf3.json`

mos.yml
timoschilling:
```suggestion
```

src/shelly_main.cpp
```diff
@@ -822,6 +825,11 @@ extern "C" bool mgos_ota_merge_fs_should_copy_file(const char *old_fs_path,
...
+      "self_test",
+      "colordata",
+      "wifiresetdata",
```
timoschilling, line 832:
```suggestion
```

mos.yml
```diff
@@ -84,18 +84,18 @@ config_schema:
...
-  - ["rgb.in_mode", "i", 1, {title: "-1 - Absent, 0 - Momentary, 1 - Toggle, 2 - Edge, 3 - Detached"}]
-  - ["rgb.in_inverted", "b", false, {title: "Invert input, set to true for normally closed input"}]
-  - ["rgb.state", "b", false, {title: "State of the light"}]
-  - ["rgb.hue", "i", 0, {title: "Hue of the light"}]
-  - ["rgb.brightness", "i", 100, {title: "Brightness of the light"}]
-  - ["rgb.saturation", "i", 100, {title: "Saturation of the light"}]
-  - ["rgb.initial_state", "i", 3, {title: "Initial state on power-on: 0 - off, 1 - on, 2 - restore last state, 3 - matches input if in toggle mode, otherwise off"}]
-  - ["rgb.auto_off", "b", false, {title: "Whether the switch should automatically turn OFF after turning ON"}]
-  - ["rgb.auto_off_delay", "d", 0, {title: "Delay for automatically turning OFF, in seconds"}]
+  - ["lb", "o", {title: "Light bulb settings", abstract: true}]
```
timoschilling, line 87: I decided to name it "lb" for light bulb to be compatible for upcoming changes for 4 chanel support and support for vintage, duo and color bulb.

src/shelly_main.cpp
rojer: would be interesting to know which exact settings cause it. let's do it separately.

rojer (commented):
ok, let's go with this

rojer:
congrats and huge thanks for doing this, Timo!

src/shelly_main.cpp
timoschilling: It was something with color mode, that was the only key in conf3

PR #567, opened 2021-03-15
improve log access in local dev mode and always show debug link

timoschilling:
Changes:
1. With this PR the debug link also works in the local frontend dev mode.
2. I see no longer a need to hide the debug link on load as it's always shown after load.


timoschilling:
@rojer any reasons not to merge this?

rojer:
sorry, just fell of the radar.
merged, thanks!

Issue #443, opened 2021-01-27
UI: Think about a scaleable, future-proof Framework for Frontend


col-tom:
The project is continuously growing and therefore also the User Interface is faced with a lot of challenges in the future:

* more state (that is in sync with UI and device)
* handling state that must be persisted on the device and therefore handling success, loading state and failture (+ user feedback)
* different UI controls / forms that depend on other state
* form validation
* async data
* mobile view / responsive
* ...

Currently the Javascript code is in one big file and not really easy to maintain, also tests for the frontend is missing.
Personally I would like to have a lightweight framework like [Preact](https://preactjs.com) or similar (component based development).

This issue should serve as a starting point for ideas and discussions.

rojer:
one thing to keep in mind is that our space budget is very tight - current filesystem size is 256K but will likely be reduced to 128K as we are almost out of code space and will need to resize. we also need to leave some space for config, logs and other miscellaneous stuff, so realistically we don't want web content to exceed 50K in size, all told.
there is also a reason for not having multiple files: during revert to stock all the files that are not present in the stock version will be copied over and thus take up space on the stock firmware, which may have its own constraints. `index.html.gz` is present in the stock firmware so it will be overridden with the stock version.

col-tom:
ok, let's get into the discussion:

* space budget: especially for that, it's nice to have such a framework (reusable components) and with a bundler that does a lot of stuff to extremely (!) compress all the asset files (HTML, JS, CSS, ...) With a bundler we can even automatically check and warn if size is e.g. over 20k and so on
* UI development: what's the long-term solution here? IMHO the current manual appraoch isn't feasible with more developers and requriements coming in

What do you mean exactly by "having multiple files / stock firmware" ?

timoschilling:
> What do you mean exactly by "having multiple files / stock firmware" ?
We only want to have files in the firmware which are in the stock (original shelly) firmware, too. The reason is that then the stock firmware will overwrite our files if a user switches back to stock.

Other wise we can run in the case that we mess up the filesystem space and we will brick the stock firmware then.

But if we bundle our assets into the index.html like we do it now, we are fine. Then we only need to handle with the size limit.

Have you seen that I have improved the development system, you can now open the `fs_src/index.html` local in your browser and connect them with a shelly. That means you can simply edit, safe and reload to view the changes.

Some days ago I reduced the fs size by ~4k and we now by ~13k, so we have some new space for preact, I think.

[...]

PR #573, opened 2021-03-17
[tech preview] reimplement UI with preact

timoschilling:
As already discussed in https://github.com/mongoose-os-apps/shelly-homekit/issues/443, the current frontend system reaches his limits. With this PR I want to show how a new system based on preact can look like.

Currently I expect that everything except the components is working. This PR is based on a commit from 12.02.2021, there are some uncovered changes in the mean time, which will be updated soon.

I assume that the footprint size of the preact version will be the same or maybe even better then the current footprint size.

@rojer If you like it and give it a 👍, I will continue the work and close the feature gap to the current UI.

Fore me the benefits are:
1. Reuse of UI components, which reduces the redundancy of html fragments
2. Simplified implicit binding between data and ui
3. more object oriented coding approach then the current spaghetti code

### DIFF
--- a/fs_src/index.html
+++ b/fs_src/index.html
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+﻿<!DOCTYPE html>
 <html lang="en">
 
 <head>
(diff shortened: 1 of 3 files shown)


fs_src/index.html
```diff
@@ -1,595 +1,551 @@
...
-      </div>
-      <div class="form-control" id="revert_to_stock_container" style="display: none">
-        <label>Revert to stock:</label>
-        <button class="btn" id="revert_btn">
-          <label><span id="revert_btn_spinner"></span>Revert</label>
-        </button>
-        <span id="revert_status"></span>
-        <div id="revert_msg" style="margin-left: 9em; display: none">Please consider reporting missing features
-          <a href="https://github.com/mongoose-os-apps/shelly-homekit/issues">on GitHub</a>.
+  <script type="module">
```
timoschilling, line 9: For now the UI script will happen inline in the html, we can change that before merging, for the initial development it's easier.

fs_src/index.html
```diff
@@ -1,595 +1,551 @@
...
+    // JS code will be extracted back to script.js if work is done
+
+    // DEV MODE
+    // import * as htm_preact_standalone from 'https://unpkg.com/htm/preact/standalone.module.js';
```
timoschilling, line 13: This is how it looks like in the local dev mode. During the build the make script will inline this js lib file into the html document and it will looks like the following line. For easier development I currently use a static permanent inlined version of preact. We can discuss how we want to handle it, including preact in the repo vs. fetching it on build.

fs_src/index.html
```diff
@@ -1,595 +1,551 @@
...
+    const htm_preact_standalone = {h: a,html: re,render: T,Component: d,createContext: U,useState: q,useReducer: B,useEffect: $,useLayoutEffect: j,useRef: z,useImperativeHandle: G,useMemo: J,useCallback: K,useContext: Q,useDebugValue: X};
+    // BUILD MODE END
+
+    const { html, Component, render, useEffect, useState } = htm_preact_standalone;
```
timoschilling, line 20: adapter between dev import and build inlined preact.

col-tom:
Thanks a lot, can you commit also the raw (readable) components itself?
Later on we should also add compression for html,css,js for sure.

timoschilling:
You mean a readable version of [this](https://github.com/mongoose-os-apps/shelly-homekit/blob/03d5c663ad6d1ca94b4dd90e4da497795e68d361/fs_src/index.html#L16)?
I can't find a readable version of it, it's size optimised merge of https://github.com/developit/htm and https://github.com/preactjs/preact, which is provided by the htm npm package. If you want to understand them, take a look on the preact and htm source code.

col-tom:
ok, I mean the single, encapsulated Preact components -> https://preactjs.com/guide/v10/components

The single components then get's compressed and bundled in the build step.

> Components represent the basic building block in Preact. They are fundamental in making it easy to build complex UIs from little building blocks.

timoschilling:
The components are there, see `fs_src/index.html` starting line ~29. Because we want a simple build pipeline they are inlined (maybe I will extract them to a .js file before finishing this PR) and will not be compressed.

fs_src/index.html
```diff
@@ -1,595 +1,551 @@
...
+      }
+
+      reconnect() {
+        this.socket = new WebSocket(`ws://${this.app.state.remoteHost}/rpc`);
+        this.socket.onerror = () => {}
+        this.socket.onopen = () => location.reload();
+      }
+
+      send(method, params = {}, callback = null) {
+        let requestId = this.requestId++;
```
rojer, line 438: id 0 is not a valid request id, should start at 1 (or better yet, a random integer).

fs_src/index.html
```diff
@@ -1,595 +1,551 @@
...
+      getInfo() {
+        this.send("Shelly.GetInfo");
+      }
+
+      setConfig(config, completeHandler) {
+        this.send("Shelly.SetConfig", { config }, completeHandler);
+      }
+
+      onMessage({ data }) {
+        let { id, result } = JSON.parse(data);
```
rojer, line 453: need to handle errors (when request returns an error with code and message instead of result)

rojer (commented):
so, as someone who is unfamiliar with preact, this looks pretty much the same to me in terms of readability.
the ui is not functional enough to tell what the UX is like. i'll note that we are adding about as much code as we are removing (in terms of lines, so i'm the entire preact is one line).

what will this change enable that wis impossible/cumbersome today?

sergioramos:
@timoschilling have you considered something like [esbuild](https://esbuild.github.io/) to bundle and minify the script - and that way save precious space?

rojer:
minified scripts are a pain to debug.
let's not concern ourselves with optimization at this point, we are not running out of space yet, we can afford readable code.
i like Timo's approach here - include the library as a minified blob and write our app code as non-minified.

col-tom:
> minified scripts are a pain to debug.

Therefore we could use auto-build source-maps (then in the browser it looks like unminified - good and is easy to debug)

timoschilling:
I agree with @rojer in general. Pleas keep in mind that the most work here is done by people how are no frontend devs in there day job. There are reasons for not using things like source maps. I tested this PR with minified but in the end there was no real benefit while the gzip brings more then minified. 

Please don’t let us discuss the pipeline line in this PR. This PR is about a new UI not about a build pipeline, when you want a new build pipeline open a issue for that and discussion it there.

timoschilling:
@rojer At the moment the line count is similar, because of some tool functions. I will write an explicit compare example how some thing looks in the old and on the new way, in the next days.

Andreyco:
What is storage capacity in Shelly devices?

rojer:
@Andreyco FS size is 256K atm, but not all of it can be used for files. also, it may be reduced to 128K in the future. in practice, we can afford to spend about 50K for the UI.

timoschilling:
Closed because completely out of date, maybe redo it in the future.

PR #581, opened 2021-03-19
Flashscript: Version 2.6.2

andyblac:
fix crash for stock now sending '{'status': 'Updating, please wait...'}' in '/rpc/Shelly.GetInfo' during update.


andyblac:
i'll find a better way, this was just a quick fix.

PR #590, opened 2021-03-23
[Flashscript] Version 2.7.0

### LINKED ISSUE #589: regular expressions strings
Only a minor aesthetic, yet perhaps can be more

    hostcheck = re.search(r"\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}", host)

shouldn't regexp string be prefixed by **r** so the backslashes are preserved?

Another:

     logger.trace(f"\n…

inserts a newline before the string, doesn't it?

andyblac:
- move run block into Main class, move config, device_scan, manual_hosts into own functions, better handling of webserver shutdown.
- add support for RGBW2 (both stock color modes).
- add support for ShellyMotion (list info only, must use manual IP's and use --timeout 240, as motion needs longer to wake and be detected)
- add -dn | --device-name filter to limit scan to include term.
- add '-f | --flash', '--app-version', '--timeout'
- add battery info for those devices that support it.
- python formatting and cleanup, more optimisations
- add regex literal to re.search lines, suggestion by juanfal thanks. fixes #589
- fix keyboard interrupt for manual hosts.
- if device on stock, but not on latest version, nolonger show latest shock version info, but HomeKit version available.

### DIFF
--- a/tools/flash-shelly.py
+++ b/tools/flash-shelly.py
@@ -615,21 +615,22 @@ class Main:
     return major, minor, patch, variant, var_seq
 
   def is_newer(self, v1, v2):
-    logger.trace(f"is: {v1}  newer than: {v2}")
     vi1 = self.parse_version(v1)
     vi2 = self.parse_version(v2)
     if vi1[0] != vi2[0]:
-      return vi1[0] > vi2[0]
+      value = vi1[0] > vi2[0]
     elif vi1[1] != vi2[1]:
-      return vi1[1] > vi2[1]
+      value = vi1[1] > vi2[1]
     elif vi1[2] != vi2[2]:
-      return vi1[2] > vi2[2]
+      value = vi1[2] > vi2[2]
     elif vi1[3] != vi2[3]:
-      return True
+      value = True
     elif vi1[4] != vi2[4]:
-      return vi1[4] > vi2[4]
+      value = vi1[4] > vi2[4]
     else:
-      return False
+      value = False
+    logger.trace(f"is {v1} newer than {v2}: {value}")
+    return value
 
   def write_network_type(self, device_info):
     logger.debug(f"{PURPLE}[Write Network Type]{NC}")
--- a/tools/flash-shelly.py
+++ b/tools/flash-shelly.py
@@ -235,13 +235,8 @@ class Device:
     self.variant = main.variant
     self.version = main.version
     self.local_file = main.local_file
-    self.stock_model = None
     self.fw_version = None
     self.fw_type_str = None
-    self.app = None
-    self.model = None
-    self.color_mode = None
-    self.device_id = None
     self.flash_fw_version = '0.0.0'
     self.flash_fw_type_str = None
     self.download_url = None
@@ -428,16 +423,16 @@ class Device:
         self.download_url = 'local'
       else:
         self.download_url = f'http://{local_ip}:{webserver_port}/{self.local_file}'
-      if self.is_stock() and self.stock_model == 'SHRGBW2' and self.color_mode is not None:
-        m_model = f"{self.app}-{self.color_mode}"
+      if self.is_stock() and self.info.get('stock_model') == 'SHRGBW2' and self.info.get('color_mode'):
+        m_model = f"{self.info.get('app')}-{self.info.get('color_mode')}"
       else:
-        m_model = self.app
+        m_model = self.info.get('app')
       if m_model != manifest_name:
         self.flash_fw_version = '0.0.0'
         self.download_url = None
       return True
     else:
-      logger.debug(f"File does not exist")
+      logger.info(f"File does not exist.")
       self.flash_fw_version = '0.0.0'
       self.download_url = None
       return False
@@ -457,17 +452,19 @@ class Device:
           re_search = i[0]
         if re.search(re_search, self.fw_version):
           self.flash_fw_version = i[1].get('version', '0.0.0')
-          self.download_url = i[1].get('urls', {}).get(self.model)
+          self.download_url = i[1].get('urls', {}).get(self.info.get('model'))
           break
     else:
       self.flash_fw_version = self.version
-      self.download_url = f'http://rojer.me/files/shelly/{self.version}/shelly-homekit-{self.model}.zip'
+      self.download_url = f"http://rojer.me/files/shelly/{self.version}/shelly-homekit-{self.info.get('model')}.zip"
 
   def update_stock(self, release_info=None):
     self.flash_fw_type_str = 'Stock'
+    stock_model = self.info.get('stock_model')
+    color_mode = self.info.get('color_mode')
     logger.debug(f"Mode: {self.fw_type_str} To {self.flash_fw_type_str}")
     if not self.version:
-      stock_model_info = release_info.get('data', {}).get(self.stock_model)
+      stock_model_info = release_info.get('data', {}).get(stock_model)
       if self.variant == 'beta':
         self.flash_fw_version = self.parse_stock_version(stock_model_info.get('beta_ver', '0.0.0'))
         self.download_url = stock_model_info.get('beta_url')
@@ -476,10 +473,10 @@ class Device:
         self.download_url = stock_model_info.get('url')
     else:
       self.flash_fw_version = self.version
-      self.download_url = f'http://archive.shelly-tools.de/version/v{self.version}/{self.stock_model}.zip'
-    if self.stock_model == 'SHRGBW2':
-      if self.download_url and not self.version and self.color_mode and not self.local_file:
-        self.download_url = self.download_url.replace('.zip', f'-{self.color_mode}.zip')
+      self.download_url = f'http://archive.shelly-tools.de/version/v{self.version}/{stock_model}.zip'
+    if stock_model == 'SHRGBW2':
+      if self.download_url and not self.version and color_mode and not self.local_file:
+        self.download_url = self.download_url.replace('.zip', f'-{color_mode}.zip')
 
 
 class HomeKitDevice(Device):
@@ -488,11 +485,7 @@ class HomeKitDevice(Device):
       return False
     self.fw_type_str = 'HomeKit'
     self.fw_version = self.info.get('version', '0.0.0')
-    self.model = self.info.get('model', self.shelly_model(self.info.get('app', {})[0]))
-    self.stock_model = self.info.get('stock_model')
-    self.app = self.info.get('app', self.shelly_model(self.stock_model)[1])
-    self.device_id = self.info.get('device_id')
-    self.color_mode = self.info.get('mode')
+    self.info['color_mode'] = 'color'
     return True
 
   def flash_firmware(self):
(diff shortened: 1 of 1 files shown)


andyblac:
don't merge yet, i'll mark for review once ready.

tools/flash-shelly.py
```diff
@@ -269,6 +269,18 @@ def get_device_info(self, is_flashing=False):
...
+    if self.fw_type == 'homekit':
+      return True
+    else:
+      return False
+
+  def is_stock(self):
+    if self.fw_type == 'stock':
+      return True
+    else:
+      return False
```
timoschilling, line 282:
```suggestion
  def is_homekit(self):
    return self.fw_type == 'homekit'

  def is_stock(self):
    return self.fw_type == 'stock'
```

tools/flash-shelly.py
```diff
@@ -290,19 +302,19 @@ def set_force_version(self, version):
   def get_current_version(self, is_flashing=False): # used when flashing between formware versions.
     if not self.get_device_info(is_flashing):
       return None
-    if self.fw_type == 'homekit':
+    if self.is_homekit:
```
timoschilling, line 305:
```suggestion
    if self.is_homekit():
```

tools/flash-shelly.py
andyblac: thx

tools/flash-shelly.py
andyblac: thx

rojer:
i think saving and deleting config should not be handled by the script. if user wants to provide a config, they should write one. same, if a user wants to get rid of the config, there's `rm` or `del`.
i'm still not convinced that config is even needed, for average user.
i'm sure advanced users are capable of storing different invocations of the script.
@timoschilling you requested it, can you elaborate?

andyblac:
> i think saving and deleting config should not be handled by the script. if user wants to provide a config, they should write one. same, if a user wants to get rid of the config, there's `rm` or `del`.
> i'm still not convinced that config is even needed, for average user.
> i'm sure advanced users are capable of storing different invocations of the script.
> @timoschilling you requested it, can you elaborate?

cant be done ALL flag have to be included in config, not option you want, as it use configparser it sets new default from the config. i can remove delete option np though.

andyblac:
sample of a config section
```
[motionsensors]
mode = keep
info_level = 3
fw_type_filter = all
model_type_filter = all
device_name_filter = all
do_all = False
quiet_run = False
list = True
exclude = 
dry_run = False
silent_run = False
version = 
variant = 
local_file = 
hap_setup_code = 
network_type = 
ipv4_ip = 
ipv4_gw = 
ipv4_mask = 
ipv4_dns = 
verbose = 3
timeout = 360
log_filename = 
reboot = False
hosts = 192.168.3.20 192.168.3.43 192.168.3.46 192.168.3.47
```

rojer:
> cant be done, as ALL flags have to be included in config, not option you want

and yet it has to be. otherwise every new option will break existing user configs.
if configparser can't do defaults (i find it hard to believe), don't use configparser, use something else.
YAML is a nice, user friendly config language but unfortunately an extra dependency.
JSON is in the standard library so it's tempting to use but it's a terrible choice for configs, let's not.

rojer:
but back to the question: why do we need this at all?

andyblac:
this is code

```
    parser.set_defaults(mode=config.get(args.config, 'mode'), info_level=config.getint(args.config, 'info_level'), fw_type_filter=config.get(args.config, 'fw_type_filter'),
                        model_type_filter=config.get(args.config, 'model_type_filter'), device_name_filter=config.get(args.config, 'device_name_filter'),
                        do_all=config.getboolean(args.config, 'do_all'), quiet_run=config.getboolean(args.config, 'quiet_run'), list=config.getboolean(args.config, 'list'),
                        exclude=config.get(args.config, 'exclude'), dry_run=config.getboolean(args.config, 'dry_run'), silent_run=config.getboolean(args.config, 'silent_run'),
                        version=config.get(args.config, 'version'), variant=config.get(args.config, 'variant'), local_file=config.get(args.config, 'local_file'),
                        hap_setup_code=config.get(args.config, 'hap_setup_code'), network_type=config.get(args.config, 'network_type'), ipv4_ip=config.get(args.config, 'ipv4_ip'),
                        ipv4_gw=config.get(args.config, 'ipv4_gw'), ipv4_mask=config.get(args.config, 'ipv4_mask'), ipv4_dns=config.get(args.config, 'ipv4_dns'),
                        verbose=config.getint(args.config, 'verbose'), timeout=config.get(args.config, 'timeout'), log_filename=config.get(args.config, 'log_filename'),
                        reboot=config.getboolean(args.config, 'reboot'), hosts=config.get(args.config, 'hosts').split())
```

andyblac:
> but back to the question: why do we need this at all?

for example, he shelly motions are not found by the scanner bue to the way they are to preserve battery, the only way to wake the devcie for web access or rpc is to connect to it via port 80 on the ip, hostname also do not work, so when using the script with --config {profile name} you can load setting required to handle them in one go, without having to remember their IP's
```
andrews-mbp-wifi:tools andrewblackburn$ ./flash-shelly.py --timeout 360 -l -i3 -m=keep 192.168.3.46 192.168.3.47 192.168.3.20 192.168.3.43 --save-config motionsensors
Saving configuration section motionsensors to .flashrc
andrews-mbp-wifi:tools andrewblackburn$ ./flash-shelly.py --config motionsensors
Reading configuration section motionsensors from .flashrc
Looking for Shelly devices...

Host: http://192.168.3.46
Device Name: Hall Motion Sensor
Model: ShellyMotion
Device ID: shellymotionsensor-60A4239A69B0
SSID: ABFi-IoT-2.4
IP: 192.168.3.46
RSSI: -42
Up Time: 1 day, 12:21:37
Battery: 91%
Firmware: Stock 1.1.0 ✔

Host: http://192.168.3.47
Device Name: Landing Motion Sensor
Model: ShellyMotion
Device ID: shellymotionsensor-60A4239368A2
SSID: ABFi-IoT-2.4
IP: 192.168.3.47
RSSI: -33
Up Time: 12 days, 4:14:50
Battery: 94%
Firmware: Stock 1.1.0 ✔

Host: http://192.168.3.20
Device Name: Back Garden Motion Sensor
Model: ShellyMotion
Device ID: shellymotionsensor-60A423976642
SSID: ABFi-IoT-2.4
IP: 192.168.3.20
RSSI: -41
Up Time: 10:55:41
Battery: 95%
Firmware: Stock 1.1.0 ✔

Host: http://192.168.3.43
Device Name: Front Garden Motion Sensor
Model: ShellyMotion
Device ID: shellymotionsensor-60A4239752B4
SSID: ABFi-IoT-2.4
IP: 192.168.3.43
RSSI: -55
Up Time: 14:43:22
Battery: 96%
Firmware: Stock 1.1.0 ✔

Devices found: 4 Upgradeable: 0
```

rojer:
how is this different from having `motionsensors.sh` with `./flash-shelly.py --timeout 360 -l -i3 -m=keep 192.168.3.46 192.168.3.47 192.168.3.20 192.168.3.43` ?

andyblac:
> how is this different from having `motionsensors.sh` with `./flash-shelly.py --timeout 360 -l -i3 -m=keep 192.168.3.46 192.168.3.47 192.168.3.20 192.168.3.43` ?

you can't overrule those bu extra command options
```motionsensors.sh -i2`

**Comment by rojer** [comment]
sure i can:

`./flash-shelly.py --timeout 360 -l -i3 -m=keep "$@" 192.168.3.46 192.168.3.47 192.168.3.20 192.168.3.43`

**Comment by andyblac** [comment]
> > cant be done, as ALL flags have to be included in config, not option you want
> 
> and yet it has to be. otherwise every new option will break existing user configs.
> if configparser can't do defaults (i find it hard to believe), don't use configparser, use something else.
> YAML is a nice, user friendly config language but unfortunately an extra dependency.
> JSON is in the standard library so it's tempting to use but it's a terrible choice for configs, let's not.

is it not only down file format, but argparser. you have to set new defaults from loaded config however you save / load it.

**Comment by andyblac** [comment]
> sure i can:
> 
> `./flash-shelly.py --timeout 360 -l -i3 -m=keep "$@" 192.168.3.46 192.168.3.47 192.168.3.20 192.168.3.43`

try it i bet it complains about 2 lots of same option ie `i3 i2`

**Comment by rojer** [comment]
nope, it just uses the last value

**Comment by andyblac** [comment]
tbh i do either way, but personally i dont see an issue if it not default.

**Comment by rojer** [comment]
> i dont see an issue if it not default.

sorry, i can;t parse that :) can you rephrase?

**Comment by andyblac** [comment]
> > i dont see an issue if it not default.
> 
> sorry, i can;t parse that :) can you rephrase?

i do not see use config's being issue, as long as option 2 in voting is not default.

**Comment by rojer** [comment]
the issue is of extra complexity - in my opinion it is not warranted.
features have cost, they need to be documented, maintained, tested and re-tested, etc.
we just don't need this, imo.

**Comment by andyblac** [comment]
> the issue is of extra complexity - in my opinion it is not warranted.
> features have cost, they need to be documented, maintained, tested and re-tested, etc.
> we just don't need this, imo.

i dont see how configs add's extra complexity, it is simply just setting new defaults. ?

**Comment by rojer** [comment]
it's code complexity. a fair bit of code is dedicated to parsing config and getting values from it. every flag has to be duplicated there, you need to remember that when adding flags, failing to do that will break config functionality, silently.
plus, in its current form it's unusable (due to requiring having all the flags set), this will need to change.

so, my vote is against having the config at all, we just don't need it at this time.

**Comment by andyblac** [comment]
> plus, in its current form it's unusable (due to requiring having all the flags set), this will need to change.
> 
> so, my vote is against having the config at all, we just don't need it at this time.

i dont how you IF you want to keep command line prioity.

**Comment by rojer** [comment]
i'm sure there's a way but i don't even want to spend time looking because it's irrelevant.
i think the whole things should go, there's no point in even spending time on it.
i bet at this point more time has been spent between the three of us developing and discussing it than it would have ever saved.

i'll let @timoschilling put forward his argument though, maybe i'm missing something.

**Comment by timoschilling** [comment]
@rojer sure your way works as well, the config way is just more comfortable and a quite common pattern.

**Comment by rojer** [comment]
> @rojer sure your way works as well, the config way is just more comfortable and a quite common pattern.

i know configs are common, just not for this.
so, the final verdict is: no config, for now.

**Comment by andyblac** [comment]
ok i'll remove.

File: tools/flash-shelly.py [hunk 3]
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 """
-Copyright (c) 2021 Andrew Blackburn & Deomid "rojer" Ryabkov
+Copyright (c) 2021 Andrew Blackburn, Deomid "rojer" Ryabkov & Timo Schilling
```
rojer, line 3: "Shelly-HomeKit Contributors"

andyblac:
@rojer @timoschilling please give this a test mainly 

flashing from factory stock to hk
flashing RGBW2 from stock to hk, both white / color modes
reverting from hk to stock.

if any issues found please run again with `--log-file debug.log` flag and post file.

once were happy, i'll start cleaning up verbose log info.

thx

andyblac:
@rojer ready for review.

PR #596, opened 2021-03-27
Rework HSV to RGB calculation

jobl1978:
Incoming h, s, v values should not be used as a reference. References should only be used if method will change them like out-params. There is no performance advantage in this case, because float = 32bit and references are also 32bit. Incoming parameters (larger than 32bit) should use "const 'datatype' &".

The method doesn't change class members, in this case the method should be const.

I also fixed the HSV to RGB calculation, it wasn't working correct

HSV input = H (0-359) S (0-1) V (0-1)
RGB output = R (0-1) G (0-1) B (0-1)

I think it would also be better to declare structs for RGB and HSV values to improve performance. I didn't change that, because there is already a class named RGB in use. In future it should be possible to make a smooth animated transition between 2 colors. In this case this conversion function will be called very often.

Example:
struct HSV { float h; float s; float v; };
struct RGB { float r; float g; float b; };
RGB HSVtoRGB(const HSV &hsv) const;

### DIFF
--- a/src/shelly_hap_rgb.cpp
+++ b/src/shelly_hap_rgb.cpp
@@ -24,8 +24,6 @@
 
 #include "mgos_hap_accessory.hpp"
 
-#include <math.h>
-
 namespace shelly {
 namespace hap {
 
@@ -145,50 +143,51 @@ void RGB::HSVtoRGB(float h, float s, float v, float &r, float &g,
   if (s == 0.0) {
     // if saturation is zero than all rgb hannels same as brightness
     r = g = b = v;
-  } else {
-    int i = static_cast<int>(h * 6);
-    float f = (h * 6.0f - i);
-    float p = v * (1.0f - s);
-    float q = v * (1.0f - f * s);
-    float t = v * (1.0f - (1.0f - f) * s);
+    return;
+  }
 
-    switch (i % 6) {
-      case 0:
-        r = v;
-        g = t;
-        b = p;
-        break;
+  int i = static_cast<int>(h * 6);
+  float f = (h * 6.0f - i);
+  float p = v * (1.0f - s);
+  float q = v * (1.0f - f * s);
+  float t = v * (1.0f - (1.0f - f) * s);
 
-      case 1:
-        r = q;
-        g = v;
-        b = p;
-        break;
+  switch (i % 6) {
+    case 0:
+      r = v;
+      g = t;
+      b = p;
+      break;
 
-      case 2:
-        r = p;
-        g = v;
-        b = t;
-        break;
+    case 1:
+      r = q;
+      g = v;
+      b = p;
+      break;
 
-      case 3:
-        r = p;
-        g = q;
-        b = v;
-        break;
+    case 2:
+      r = p;
+      g = v;
+      b = t;
+      break;
 
-      case 4:
-        r = t;
-        g = p;
-        b = v;
-        break;
+    case 3:
+      r = p;
+      g = q;
+      b = v;
+      break;
 
-      case 5:
-        r = v;
-        g = p;
-        b = q;
-        break;
-    }
+    case 4:
+      r = t;
+      g = p;
+      b = v;
+      break;
+
+    case 5:
+      r = v;
+      g = p;
+      b = q;
+      break;
   }
 }
 
--- a/src/shelly_hap_rgb.cpp
+++ b/src/shelly_hap_rgb.cpp
@@ -140,7 +140,7 @@ Status RGB::Init() {
 
 void RGB::HSVtoRGB(float h, float s, float v, float &r, float &g, float &b) {
   if (s == 0.0) {
-    // if saturation is zero than all rgb hannels same as brightness
+    // if saturation is zero than all rgb channels same as brightness
     r = g = b = v;
     return;
   }
--- a/src/shelly_hap_rgb.cpp
+++ b/src/shelly_hap_rgb.cpp
@@ -138,7 +138,8 @@ Status RGB::Init() {
   return Status::OK();
 }
 
-void RGB::HSVtoRGB(float &h, float &s, float &v, float &r, float &g, float &b) {
+void RGB::HSVtoRGB(float h, float s, float v,
+                   float &r, float &g, float &b) const {
   if (s == 0.0) {
     r = g = b = 1.0;
   }
--- a/src/shelly_hap_rgb.cpp
+++ b/src/shelly_hap_rgb.cpp
@@ -146,36 +146,48 @@ void RGB::HSVtoRGB(float h, float s, float v, float &r, float &g,
     // if saturation is zero than all rgb hannels same as brightness
     r = g = b = v;
   } else {
-    float h1 = fmod(h, 360.0f);  // jail it to 0-359°
+    float h1 = fmod(h, 360.0f);  // jail hue into 0-359°
     float c = v * s;
     float h2 = h1 / 60.0f;
     float x = c * (1.0f - fmod(h2, 2.0f) - 1.0f);
     float m = v - c;
 
-    if (h1 < 60.0f) {
-      r = c;
-      g = x;
-      b = 0;
-    } else if (h1 < 120.0f) {
-      r = x;
-      g = c;
-      b = 0;
-    } else if (h1 < 180.0f) {
-      r = 0;
-      g = c;
-      b = x;
-    } else if (h1 < 240.0f) {
-      r = 0;
-      g = x;
-      b = c;
-    } else if (h1 < 300.0f) {
-      r = x;
-      g = 0;
-      b = c;
-    } else if (h1 < 360.0f) {
-      r = c;
-      g = 0;
-      b = x;
+    switch (static_cast<int>(h2)) {
+      case 0:
+        r = c;
+        g = x;
+        b = 0;
+        break;
+
+      case 1:
+        r = x;
+        g = c;
+        b = 0;
+        break;
+
+      case 2:
+        r = 0;
+        g = c;
+        b = x;
+        break;
+
+      case 3:
+        r = 0;
+        g = x;
+        b = c;
+        break;
+
+      case 4:
+        r = x;
+        g = 0;
+        b = c;
+        break;
+
+      case 5:
+        r = c;
+        g = 0;
+        b = x;
+        break;
     }
 
     r += m;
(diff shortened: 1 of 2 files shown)


timoschilling:
Your calculation doesn’t work! It has something to do how you deal with v. Will explain it later, sorry im currently on the way.

timoschilling:
Can you explain the problems with the calculation.

jobl1978:
I tried and verified It out here -> https://www.peko-step.com/en/tool/hsvrgb_en.html

The new implementation outputs now the exact same values as there

If saturation is zero than every RGB channel always represents the brightness (V) and not hardcoded 1.0.
A saturation (S) of 0% and a brightness (V) of 50% (0.5) -> RGB(0.5, 0.5, 0.5) -> RGB(128, 128, 128)

Your code
if (s == 0.0) {
    r = g = b = 1.0;
  }
////////////// no else?
  int i = int(h * 6.0);
  float f = (h * 6.0) - i;
  float p = v * (1.0 - s);
  float q = v * (1.0 - s * f);
  float t = v * (1.0 - s * (1.0 - f));

andyblac:
> If saturation is zero than every RGB channel always represents the brightness (V) and not hardcoded 1.0.
> A saturation (S) of 0% and a brightness (V) of 50% (0.5) -> RGB(0.5, 0.5, 0.5) -> RGB(128, 128, 128)

sorry but i disagree,
then how can we get get a 50% brightness of 100% Blue ? and not a 50% Brigtness of 50% blue

this is this issue i reported whilst timo and I where testing.

jobl1978:
https://en.wikipedia.org/wiki/HSL_and_HSV#HSL_to_RGB

andyblac:
> https://en.wikipedia.org/wiki/HSL_and_HSV#HSL_to_RGB

please answer the question.

andyblac:
i'll wait for timo to respond.

jobl1978:
It will always have an impact to the color if you change one of the HSV channels

andyblac:
> It will always have an impact to the color if you change one of the HSV channels

well currently i can control brightness, without it changing the colour, imo this is how it should be.

jobl1978:
I did not try it out with real hardware, I only tested the math with the common conversion method that is used for HSV -> RGB transformations

andyblac:
> I did not try it out with real hardware, I only tested the math with the common conversion method that is used for HSV -> RGB transformations

all i am trying to say is, if i only want the blue diode at 100% R / G at 0% (ie OFF) and if i change brightness it should not alter colour of any of the diodes R, G, B. blue should stay at 100% and R G at 0% (OFF)

timoschilling:
The current implementation is based of pythons (core) color system lib, so I expect that it is right.

As @andyblac already says, there is a difference in calculating color for a RGB stripe and calculating a color for a display.

I will wire up two shellys and make some compare tests with your calculation and the current calculation.

@jobl1978 from the mathematic point of view you are definitely right!

jobl1978:
That is wrong!!!!

> > I did not try it out with real hardware, I only tested the math with the common conversion method that is used for HSV -> RGB transformations
> 
> all i am trying to say is, if i only want the blue diode at 100% R / G at 0% (ie OFF) and if i change brightness it should not alter colour of any of the diodes R, G, B. blue should stay at 100% and R G at 0% (OFF)
> 
> i should also beable to have B 100%, R 0%, G 0%, at 10% brightness

That is wrong and never possible! 100% Blue means full voltage output to the B channel, but if you decrease to 10% brightness than the voltage must drop down to 10% of that output

andyblac:
> That is wrong!!!!
> 
> > > I did not try it out with real hardware, I only tested the math with the common conversion method that is used for HSV -> RGB transformations
> > 
> > 
> > all i am trying to say is, if i only want the blue diode at 100% R / G at 0% (ie OFF) and if i change brightness it should not alter colour of any of the diodes R, G, B. blue should stay at 100% and R G at 0% (OFF)
> > i should also beable to have B 100%, R 0%, G 0%, at 10% brightness
> 
> That is wrong and never possible! 100% Blue means full voltage output to the B channel, but if you decrease to 10% brightness than the voltage must drop down to 10% of that output

well it is posible i have it now 😄

timoschilling:
> Your code
if (s == 0.0) {
r = g = b = 1.0;
}
////////////// no else?

Your right, there is a return missing, but thats only a short exit, the calculation always produces 1 for each value.

jobl1978:
Play with the sliders on this page, this is the behavior how HSV transforms to RGB
https://www.peko-step.com/en/tool/hsvrgb_en.html

andyblac:
```
352345313390 shelly_hap_rgb.cpp:187  state: on, brightness: 66, hue: 240, saturation: 100
352345317855 shelly_output.cpp:74    Output 1: OFF (HAP)
352345321969 shelly_output.cpp:74    Output 2: OFF (HAP)
352345327068 shelly_output.cpp:71    Output 3: 0.660000 (HAP)
352345677176 shelly_hap_rgb.cpp:402  State 1: on
352345684905 shelly_hap_rgb.cpp:187  state: on, brightness: 66, hue: 240, saturation: 100
352345689308 shelly_output.cpp:74    Output 1: OFF (HAP)
352345693413 shelly_output.cpp:74    Output 2: OFF (HAP)
352345698470 shelly_output.cpp:71    Output 3: 0.660000 (HAP)
352345703867 shelly_hap_rgb.cpp:425  Brightness 1: 71
352345711029 shelly_hap_rgb.cpp:187  state: on, brightness: 71, hue: 240, saturation: 100
352345715330 shelly_output.cpp:74    Output 1: OFF (HAP)
352345719407 shelly_output.cpp:74    Output 2: OFF (HAP)
352345724227 shelly_output.cpp:71    Output 3: 0.710000 (HAP)

```

timoschilling:
> That is wrong and never possible! 100% Blue means full voltage output to the B channel, but if you decrease to 10% brightness than the voltage must drop down to 10% of that output

The voltage must alway be 12V or 24V, only the duty (time percentage of the on time) will change, that's how PWM works.

jobl1978:
Ok i forgot that it is in PWM mode, but does not change what I say

timoschilling:
I have compared "my" (pythons) calculation with two different HomeKit certificated RGB bulbs and they behave the same.

I will repeat: You're right if we talk about the math point of view, but if we talk about the hardware / light way, it is different! Please test it with hardware, compare it to other (not DIY hardware) and then if you still thing the current way is wrong come back and we will continue the discussion.

jobl1978:
Can one of you test it please? I don't have a ShellyRGBW2. I would be very interested in what comes out of it

timoschilling:
> Can one of you test it please? I don't have a ShellyRGBW2. I would be very interested in what comes out of it

I will do later, I will make photo of the same stripe with the current and with your implementation.

jobl1978:
Thank you... the show must go on ;-)

andyblac:
> Can one of you test it please? I don't have a ShellyRGBW2. I would be very interested in what comes out of it

imo you are confusing rgb of a colour (display pixel) and a 3 seprate led color led. the 2 are not the same.

timoschilling:
something what you maybe not noticed:
```
if (s == 0.0) {
  r = g = b = 1.0;
```
Thats a bit wired, I know. Maybe I should it change to `r = g = b = v;`, but while we cal the function alway with `v=1` it's the same.

timoschilling:
>think it like 100% Blue pixel and display brighness 100%, reduce your display brightness does NOT change the colour of the pixel.

:+1: I thing thats the best description for the difference!

timoschilling:
https://github.com/mongoose-os-apps/shelly-homekit/blob/master/src/shelly_hap_rgb.cpp#L193

jobl1978:
Before you test my code please change initialization of 'h' before you call my version of HSVtoRGB.
My calculation needs a 'h' value between 0-359 instead of 0-1

timoschilling:
After your last changes they are behave nearly the same, but there are some bugs in yours.
For example `brightness: 85, hue: 224, saturation: 59` your code produces:
```
Output 1: 0.348500 (HAP)
Output 2: OFF (HAP)
Output 3: 0.850000 (HAP)
```
The current code produces:
```
Output 1: 0.348500 (HAP)
Output 2: 0.482233 (HAP)
Output 3: 0.850000 (HAP)
```
Following the calculation on https://www.peko-step.com/en/tool/hsvrgb_en.html `G=0` is wrong.

I still see no bugs in the current implementation, there are two glitches but for the rest it works. And why should the python std lib wrong?

The glitches are:
1. v / brightness handling
2. missing return in the guard close

rojer:
style notes:

 * HSVtoRGB should be a static method since it doesn't need instance state
 * Output parameters should be pointers, do not use non-const refs for this.

rojer:
also: `#include <cmath>` and use `std::sqrt` etc. though i don't s even see any functions used.

and finally - try to keep indentation to a minimum.
HSVtoRGB consists of a single if statement with two different branches: for zero case and everything else.
dispose with it first, then continue with the main one:
```
...
  if (foo) {
    // simple thing / error
    return;
  }
  // main branch goes here
}
```

jobl1978:
> > Can one of you test it please? I don't have a ShellyRGBW2. I would be very interested in what comes out of it
> 
> imo you are confusing rgb of a colour (display pixel) and a 3 seprate led color led. the 2 are not the same.
> 
> 0% sat = 100% colour ie FF hex
> 10% V = 25 colour value ie 19 hex
> 
> this NOT the same as
> 0% sat, 100% colour ie FF , with 10% brightness
> 
> think it like 100% Blue pixel and display brighness 100%, reduce your display brightness does NOT change the colour of the pixel.



> style notes:
> 
> * HSVtoRGB should be a static method since it doesn't need instance state
> * Output parameters should be pointers, do not use non-const refs for this.

HSVtoRGB should be static -> agree

Output parameters should always be pointers -> disagree
Reason: The interface should clearly tell the developer what is allowed and what not. A pointer tells the developer he can call it with a "nullptr" what in this case would be fatal. A Reference prevents the usage of "nullptr" for this case, because RGB will always be written.

jobl1978:
> * v / brightness handling
> * missing return in the guard close



> After your last changes they are behave nearly the same, but there are some bugs in yours.
> For example `brightness: 85, hue: 224, saturation: 59` your code produces:
> 
> ```
> Output 1: 0.348500 (HAP)
> Output 2: OFF (HAP)
> Output 3: 0.850000 (HAP)
> ```
> 
> The current code produces:
> 
> ```
> Output 1: 0.348500 (HAP)
> Output 2: 0.482233 (HAP)
> Output 3: 0.850000 (HAP)
> ```
> 
> Following the calculation on https://www.peko-step.com/en/tool/hsvrgb_en.html `G=0` is wrong.
> 
> I still see no bugs in the current implementation, there are two glitches but for the rest it works. And why should the python std lib deliver wrong results?
> 
> The glitches are:
> 
> 1. v / brightness handling
> 2. missing return in the guard close

I saw it nearly at the same time as you that some values are wrong calculated in my code :-)
I reverted it back, but I changed the usage outside the call without that hardcoded 1.0 for brightness.
Can you test the latest commit?

jobl1978:
> also: `#include <cmath>` and use `std::sqrt` etc. though i don't s even see any functions used.

I included math.h for "fmod", forgot to remove it after I did not use that function anymore ;-)

timoschilling:
The change with the brightness is good, I don't know why I done this way.

jobl1978:
> The change with the brightness is good, I don't know why I done this way.

That was that thing what confused me so hard :-D
By the way, im also a nativ German speaker Mr. Timo ;-)

timoschilling:
Please merge this into your PR
https://github.com/jobl1978/shelly-homekit/pull/3

jobl1978:
> Please merge this into your PR
> [jobl1978#3](https://github.com/jobl1978/shelly-homekit/pull/3)

Done, this rework I like also ;-)

rojer:
> Output parameters should always be pointers -> disagree
Reason: The interface should clearly tell the developer what is allowed and what not. A pointer tells the developer he can call it with a "nullptr" what in this case would be fatal. A Reference prevents the usage of "nullptr" for this case, because RGB will always be written.

ok, makes sense, no objections

rojer:
@timoschilling please let me know when it's ok to merge

rojer:
thanks!

PR #607, opened 2021-04-04
Rework OTA verification

jobl1978:
Reason for this rework is to prevent having 3 different RGBW2 firmwares if Shelly will do a OTA update from HomeKit back to Stock

### DIFF
--- a/src/Shelly1/shelly_init.cpp
+++ b/src/Shelly1/shelly_init.cpp
@@ -22,6 +22,8 @@
 
 namespace shelly {
 
+const std::set<std::string> g_compatibleFirmwareNames{"todo!!!"};
+
 void CreatePeripherals(std::vector<std::unique_ptr<Input>> *inputs,
                        std::vector<std::unique_ptr<Output>> *outputs,
                        std::vector<std::unique_ptr<PowerMeter>> *pms,
--- a/src/Shelly1/shelly_init.cpp
+++ b/src/Shelly1/shelly_init.cpp
@@ -22,7 +22,7 @@
 
 namespace shelly {
 
-const std::set<std::string> g_compatibleFirmwareNames{"todo!!!"};
+const std::set<std::string> g_compatibleFirmwareNames{"switch1"};
 
 void CreatePeripherals(std::vector<std::unique_ptr<Input>> *inputs,
                        std::vector<std::unique_ptr<Output>> *outputs,
--- a/src/Shelly1L/shelly_init.cpp
+++ b/src/Shelly1L/shelly_init.cpp
@@ -24,6 +24,8 @@
 
 namespace shelly {
 
+const std::set<std::string> g_compatibleFirmwareNames{"todo!!!"};
+
 void CreatePeripherals(std::vector<std::unique_ptr<Input>> *inputs,
                        std::vector<std::unique_ptr<Output>> *outputs,
                        std::vector<std::unique_ptr<PowerMeter>> *pms,
--- a/src/Shelly1L/shelly_init.cpp
+++ b/src/Shelly1L/shelly_init.cpp
@@ -24,7 +24,7 @@
 
 namespace shelly {
 
-const std::set<std::string> g_compatibleFirmwareNames{"todo!!!"};
+const std::set<std::string> g_compatibleFirmwareNames{"switch1l"};
 
 void CreatePeripherals(std::vector<std::unique_ptr<Input>> *inputs,
                        std::vector<std::unique_ptr<Output>> *outputs,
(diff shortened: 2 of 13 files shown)


jobl1978:
@andyblac @timoschilling @rojer Please check if all stock firmware names are correct in the set for every Shelly device

andyblac:
> @andyblac @timoschilling @rojer Please check if all stock firmware names are correct in the set for every Shelly device

here is a list https://github.com/mongoose-os-apps/shelly-homekit/blob/master/tools/flash-shelly.py#L310
STOCK MODEL, HUMAN MODEL, FW MODEL ID

jobl1978:
> > @andyblac @timoschilling @rojer Please check if all stock firmware names are correct in the set for every Shelly device
> 
> here is a list https://github.com/mongoose-os-apps/shelly-homekit/blob/master/tools/flash-shelly.py#L310
> STOCK MODEL, HUMAN MODEL, FW MODEL ID
> 
> SHRGBW2 has 3 fw id's
> `rgbw2` is base (requires you to choose mode after flashing)
> `rgbw2-white` white mode.
> `rgbw2-color` color mode.
> 
> but imo unless we are using both id modes `white` (when we add 4 channel whate support) and `color` all other modes. we could just use rgbw2-color and when reverting to just revert to color mode.

I think we should only use rgbw2 for HK, because we do not reboot and exchange the HK firmware if we switch the mode like stock firmware does. Currently we use rgw2-color.

andyblac:
> > > @andyblac @timoschilling @rojer Please check if all stock firmware names are correct in the set for every Shelly device
> > 
> > 
> > here is a list https://github.com/mongoose-os-apps/shelly-homekit/blob/master/tools/flash-shelly.py#L310
> > STOCK MODEL, HUMAN MODEL, FW MODEL ID
> > SHRGBW2 has 3 fw id's
> > `rgbw2` is base (requires you to choose mode after flashing)
> > `rgbw2-white` white mode.
> > `rgbw2-color` color mode.
> > but imo unless we are using both id modes `white` (when we add 4 channel whate support) and `color` all other modes. we could just use rgbw2-color and when reverting to just revert to color mode.
> 
> I think we should only use rgbw2 for HK, because we do not exchange the HK firmware if we switch the mode like stock firmware does. Currently we use rgw2-color.

then why are we using `-white` at all ?

jobl1978:
> > > > @andyblac @timoschilling @rojer Please check if all stock firmware names are correct in the set for every Shelly device
> > > 
> > > 
> > > here is a list https://github.com/mongoose-os-apps/shelly-homekit/blob/master/tools/flash-shelly.py#L310
> > > STOCK MODEL, HUMAN MODEL, FW MODEL ID
> > > SHRGBW2 has 3 fw id's
> > > `rgbw2` is base (requires you to choose mode after flashing)
> > > `rgbw2-white` white mode.
> > > `rgbw2-color` color mode.
> > > but imo unless we are using both id modes `white` (when we add 4 channel whate support) and `color` all other modes. we could just use rgbw2-color and when reverting to just revert to color mode.
> > 
> > 
> > I think we should only use rgbw2 for HK, because we do not exchange the HK firmware if we switch the mode like stock firmware does. Currently we use rgw2-color.
> 
> then why are we using `-white` at all ?

Because stock firmware uses it. If a OTA update is started, HK firmware checks the incoming name if it is possible or not. And stock firmware can come in with 3 different names (rgbw2, rgbw2-white, rgbw2-color)

rojer:
question: why don't we always roll back to rgbw2-color and let the user proceed from there?

timoschilling:
More convince?

rojer:
> More convince?

meaning?

timoschilling:
If we support all modes the User can switch in all directions, for me this is  more convenient

andyblac:
> If we support all modes the User can switch in all directions, for me this is more convenient

so for you it more about revert ? to both color and white, depending on what current hk mode is rgb vs wwww ?

timoschilling:
How ever let us just make one

rojer:
ok, i think this is not worth the trouble for us. let's always revert to `SHRGBW2-color.zip`. it's one of the stock configurations, user can then change to whatever they want.
it is now available at http://rojer.me/files/shelly/stock/ (i updated the [script](http://rojer.me/files/shelly/stock/update.sh) to fetch it).

andyblac:
> ok, i think this is not worth the trouble for us. let's always revert to `SHRGBW2-color.zip`. it's one of the stock configurations, user can then change to whatever they want.
> it is now available at http://rojer.me/files/shelly/stock/ (i updated the [script](http://rojer.me/files/shelly/stock/update.sh) to fetch it).

that's what i do in script.

PR #608, opened 2021-04-04
Add RGBW support

jobl1978:
Now its possible to switch the mode between RGB and RGBW

### DIFF
--- a/mos.yml
+++ b/mos.yml
@@ -96,6 +96,7 @@ config_schema:
   - ["lb.initial_state", "i", 3, {title: "Initial state on power-on: 0 - off, 1 - on, 2 - restore last state, 3 - matches input if in toggle mode, otherwise off"}]
   - ["lb.auto_off", "b", false, {title: "Whether the switch should automatically turn OFF after turning ON"}]
   - ["lb.auto_off_delay", "d", 0, {title: "Delay for automatically turning OFF, in seconds"}]
+  - ["lb.device_type", "i", 0, {title: "0 - RGB, 1 - RGBW"}]
 
   # Preserved from stock.
   - ["k_apparent", "i", 1750000, {}]
--- a/mos.yml
+++ b/mos.yml
@@ -31,7 +31,7 @@ config_schema:
   - ["wifi.sta_all_chan_scan", true]
 
   - ["shelly.name", "s", "", {title: "User-facing device name"}]  # Note: will be overridden by cfg 2 -> 3 migration.
-  - ["shelly.mode", "i", 0, {title: "System mode: 0 - default, 1 - roller shutter (if supported), 2 - garage door opener (if supported)"}]
+  - ["shelly.mode", "i", 0, {title: "System mode: 0 - default, 1 - roller shutter (if supported), 2 - garage door opener (if supported), 3 - RGB (if supported), 4 - RGBW (if supported)"}]
   - ["shelly.cfg_version", "i", 0, {title: "Configuration version"}]
   - ["shelly.legacy_hap_layout", "b", false, {title: "Use legacy accessory layout instead of a bridged accessory"}]
   - ["shelly.overheat_on", "i", 100, {title: "Overheat protection mode kicks in at or above this temperature"}]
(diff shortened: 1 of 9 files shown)


src/shelly_hap_rgb.cpp
```diff
@@ -326,7 +342,17 @@ Status RGB::SetState(const std::string &state_json) {
   return Status::OK();
 }
 
-void RGB::AutoOffTimerCB() {
+RGBWLight::LightMode RGBWLight::getLightMode() const {
+  switch (mgos_sys_config_get_shelly_mode()) {
```
rojer, line 346: components should not be referencing sys config directly, this breaks the abstraction boundaries.

mos.yml
```diff
@@ -31,7 +31,7 @@ config_schema:
   - ["wifi.sta_all_chan_scan", true]
 
   - ["shelly.name", "s", "", {title: "User-facing device name"}]  # Note: will be overridden by cfg 2 -> 3 migration.
-  - ["shelly.mode", "i", 0, {title: "System mode: 0 - default, 1 - roller shutter (if supported), 2 - garage door opener (if supported)"}]
+  - ["shelly.mode", "i", 0, {title: "System mode: 0 - default, 1 - roller shutter (if supported), 2 - garage door opener (if supported), 3 - RGB (if supported), 4 - RGBW (if supported)"}]
```
rojer, line 34: can Shelly RGBW2 be a roller shutter or a garage door opener? probably not.
can Shelly 2.5 be an RGBW PWM controller? definitely ot.

RGB is not a mode, it should be the only component instantiated for RGBW2 device in shelly_init, and no other device can be using it.
RGBW is a mode for the RGB component, it should be a boolean option.

rojer (changes requested):

mos.yml
timoschilling: We will add later a 4xW mode and RGB will be renamed  to RGB+W so that you can use RGB as now and the W as a different channel, so the mode creates 1, 2 or 3 components. I unterstand your meaning at the moment, but on the long term mode is the right place.

timoschilling: Fun fact: RGBW2 as garage door opener works, just change the pins and flash the 2.5 firmware onto the RGBW2

src/shelly_hap_rgb.cpp
```diff
@@ -101,113 +103,126 @@ Status RGB::Init() {
...
        cfg_->brightness, cfg_->hue, cfg_->saturation));
 
-  float r = 0, g = 0, b = 0;
+  RGBW rgbw{0.0f, 0.0f, 0.0f, 0.0f};
 
   if (cfg_->state) {
-    float h = cfg_->hue / 360.0f;
-    float s = cfg_->saturation / 100.0f;
-    float v = cfg_->brightness / 100.0f;
+    HSV hsv;
```
rojer, line 214:
```c
HSV hsv = {
    .h = ...,
    .s = ...,
    .v = ...,
};
```
whether c or c++, always use designated initializers to initialize structs.
this guarantees that new members (if ever added) are initialized with 0.

src/shelly_hap_rgb.cpp
```diff
@@ -101,113 +103,126 @@ Status RGB::Init() {
...
 }
 
-void RGB::SetOutputState(const char *source) {
+void RGBWLight::SetOutputState(const char *source) {
   LOG(LL_INFO,
       ("state: %s, brightness: %i, hue: %i, saturation: %i", OnOff(cfg_->state),
        cfg_->brightness, cfg_->hue, cfg_->saturation));
 
-  float r = 0, g = 0, b = 0;
+  RGBW rgbw{0.0f, 0.0f, 0.0f, 0.0f};
```
rojer, line 211: `RGBW rgbw{}`;

avoid aggregate initalizers, they are silent bugs waiting to happen - if fields of the struct are reshuffled, for example.
use designated or, if initialized with zeroes, just omit the initializer body (or use `{0} in C`).

rojer (changes requested):

mos.yml
rojer: ok, makes sense.

src/shelly_hap_rgb.cpp
jobl1978: Normally I use 

RGBW { .r = 1.0f, .g = 1.0f, .b = 10f, .w = 1.0f };

this should be safe, but for zero {0} would be better I agree

rojer: > RGBW rgbw{ .r = 1.0f, .g = 1.0f, .b = 10f, .w = 1.0f };

this would be perfectly fine

> RGBW rgbw{ 1.0f, 1.0f, 10f, 1.0f };

this is not due to reason mentioned above.

> RGBW rgbw{ .r = 0.0f, .g = 0.0f, .b = 0.0f, .w = 0.0f };

is safe but too verbose, `RGBW rgbw{};` is nice and brief.

rojer:
nice, thanks!

PR #612, opened 2021-04-06
UI support for rgbw2 revert

timoschilling:


fs_src/script.js
```diff
@@ -1157,7 +1157,13 @@ el("revert_btn").onclick = function () {
   if(!confirm("Revert to stock firmware?")) return;
 
   el("revert_msg").style.display = "block";
-  var stockURL = `https://rojer.me/files/shelly/stock/${lastInfo.stock_model}.zip`;
+
+  var model = lastInfo.stock_model;
```
rojer, line 1161: why not just set stock model to SHRGBW2-color?

timoschilling: I expect that this will breaks the flash.py script

rojer: we'll fix it when and if it does, it's better than adding this kludge.

timoschilling: 🤔 on the first view it feels wrong, but after taking a look on the mos.yml where it is named `STOCK_FW_MODEL`, it makes more scene. But then I would like to change the name in the RPC api from `stock_model` to `stock_fw_model`.

rojer: i agree, however simply changing that will _definitely_ break the script revert for _all_ the models until new script is deployed, and since people tend to download and save it, that can be a while.
so, that will have to be done in two stages: (1) add `stock_fw_model` alongside `stock_model`, with a TODO to remove the latter (2) after some time (couple releases) remove `stock_model` and TODO.

timoschilling: @andyblac with the change suggested from @rojer we can remove the py hack.

timoschilling:
Closed because of changed requirements

fs_src/script.js
andyblac: please also bare in mind stock variables, whatever we do i want them to match, imo changing stock_model to `stock_fw_model` is find but don't add color to it. as doing so will cause issues in updating stock to stock, i need the REAL stock id.

andyblac: in stock they have `   "mode": "white",` this is color or white depending on mode, we should is set this to `color`

PR #613, opened 2021-04-07
Add transition feature for RGBW2

jobl1978:
Configurable transition time for ON/OFF and color fade

### DIFF
--- a/src/shelly_hap_rgb.cpp
+++ b/src/shelly_hap_rgb.cpp
@@ -83,14 +83,14 @@ Status RGBWLight::Init() {
   if (IsSoftReboot()) should_restore = true;
 
   if (should_restore) {
-    UpdateOnOff(cfg_->state, "init");
+    UpdateOnOff(IsOn(), "init");
   } else {
     switch (static_cast<InitialState>(cfg_->initial_state)) {
       case InitialState::kOff:
-        TurnOff("init");
+        UpdateOnOff(false, "init");
         break;
       case InitialState::kOn:
-        TurnOn("init");
+        UpdateOnOff(true, "init");
         break;
       case InitialState::kInput:
         if (in_ != nullptr &&
@@ -148,51 +148,55 @@ Status RGBWLight::Init() {
   return Status::OK();
 }
 
-void RGBWLight::HSVtoRGBW(const HSV &hsv, RGBW &rgbw) const {
-  if (hsv.s == 0.0) {
+void RGBWLight::HSVtoRGBW(RGBW &rgbw) const {
+  float h = cfg_->hue / 360.0f;
+  float s = cfg_->saturation / 100.0f;
+  float v = cfg_->brightness / 100.0f;
+
+  if (cfg_->saturation == 0) {
     // if saturation is zero than all rgb channels same as brightness
-    rgbw.r = rgbw.g = rgbw.b = hsv.v;
+    rgbw.r = rgbw.g = rgbw.b = v;
   } else {
-    // otherwise calc rgb from hsv
-    int i = static_cast<int>(hsv.h * 6);
-    float f = (hsv.h * 6.0f - i);
-    float p = hsv.v * (1.0f - hsv.s);
-    float q = hsv.v * (1.0f - f * hsv.s);
-    float t = hsv.v * (1.0f - (1.0f - f) * hsv.s);
+    // otherwise calc rgb from hsv (hue, saturation, brightness)
+    int i = static_cast<int>(h * 6);
+    float f = (h * 6.0f - i);
+    float p = v * (1.0f - s);
+    float q = v * (1.0f - f * s);
+    float t = v * (1.0f - (1.0f - f) * s);
 
     switch (i % 6) {
-      case 0:
-        rgbw.r = hsv.v;
+      case 0:  // 0° ≤ h < 60°
+        rgbw.r = v;
         rgbw.g = t;
         rgbw.b = p;
         break;
 
-      case 1:
+      case 1:  // 60° ≤ h < 120°
         rgbw.r = q;
-        rgbw.g = hsv.v;
+        rgbw.g = v;
         rgbw.b = p;
         break;
 
-      case 2:
+      case 2:  // 120° ≤ h < 180°
         rgbw.r = p;
-        rgbw.g = hsv.v;
+        rgbw.g = v;
         rgbw.b = t;
         break;
 
-      case 3:
+      case 3:  // 180° ≤ h < 240°
         rgbw.r = p;
         rgbw.g = q;
-        rgbw.b = hsv.v;
+        rgbw.b = v;
         break;
 
-      case 4:
+      case 4:  // 240° ≤ h < 300°
         rgbw.r = t;
         rgbw.g = p;
-        rgbw.b = hsv.v;
+        rgbw.b = v;
         break;
 
-      case 5:
-        rgbw.r = hsv.v;
+      case 5:  // 300° ≤ h < 360°
+        rgbw.r = v;
         rgbw.g = p;
         rgbw.b = q;
         break;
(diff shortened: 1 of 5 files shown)


src/shelly_hap_rgb.cpp
```diff
@@ -212,38 +211,84 @@ void RGBWLight::HSVtoRGBW(const HSV &hsv, RGBW &rgbw) const {
...
-    HSV hsv{.h = cfg_->hue / 360.0f,
-            .s = cfg_->saturation / 100.0f,
-            .v = cfg_->brightness / 100.0f};
+void RGBWLight::Toggle(const std::string &source) {
+  UpdateOnOff(IsOff(), source);
+}
 
-    HSVtoRGBW(hsv, rgbw);
+bool RGBWLight::IsOn() const {
+  return cfg_->state != 0;
```
rojer, line 227: cfg_->state is a bool, it should be used as bool.

come to think of it - why do we even need separate state when we have brightness?

src/shelly_hap_rgb.hpp
```diff
@@ -53,19 +53,33 @@ class RGBWLight : public Component, public mgos::hap::Service {
...
   Status SetConfig(const std::string &config_json,
                    bool *restart_required) override;
   Status SetState(const std::string &state_json) override;
 
+  void TurnOn(const std::string &source);
+  void TurnOff(const std::string &source);
+  void Toggle(const std::string &source);
+  bool IsOn() const;
+  bool IsOff() const;
+  void SetHsv(int h, int s, int v, const std::string &source);
```
rojer, line 68: abbreviations should be fully capitalized, `SetHSV`

src/shelly_hap_rgb.cpp
timoschilling: brightness and state (aka On char) are different thinks in hk, a lamp can be off and brightness 100% at the same time.

src/shelly_hap_rgb.hpp
rojer: also, we have `struct HSV`, why aren't we using it here?

src/shelly_hap_rgb.hpp
```diff
@@ -75,7 +89,12 @@ class RGBWLight : public Component, public mgos::hap::Service {
   std::vector<mgos::hap::Characteristic *> state_notify_chars_;
 
   mgos::Timer auto_off_timer_;
-  bool dirty_ = false;
+
+  mgos::Timer transition_timer_;
+  int64_t transition_start_;
```
rojer, line 94: scalar class members should have defaults to ensure they are always initialized.

src/shelly_hap_rgb.cpp
```diff
@@ -212,38 +211,84 @@ void RGBWLight::HSVtoRGBW(const HSV &hsv, RGBW &rgbw) const {
   }
 }
 
-void RGBWLight::SetOutputState(const char *source) {
-  LOG(LL_INFO,
-      ("state: %s, brightness: %i, hue: %i, saturation: %i", OnOff(cfg_->state),
-       cfg_->brightness, cfg_->hue, cfg_->saturation));
+void RGBWLight::TurnOn(const std::string &source) {
```
timoschilling, line 214: This explicit function feels over engineered, my not simply using `UpdateOnOff(true, source)` direct?
Same for `TurnOff` and `Toogle`

src/shelly_hap_rgb.hpp
```diff
@@ -75,7 +89,12 @@ class RGBWLight : public Component, public mgos::hap::Service {
   std::vector<mgos::hap::Characteristic *> state_notify_chars_;
 
   mgos::Timer auto_off_timer_;
-  bool dirty_ = false;
```
rojer, line 78: why are you removing this? this makes every SaveState call hit flash, even if there are no changes to the config.

src/shelly_hap_rgb.cpp
```diff
@@ -330,42 +374,59 @@ Status RGBWLight::SetState(const std::string &state_json) {
...
-    SetOutputState("RPC");
-  }
+  UpdateOnOff(state, "RPC");
+  SetHsv(hue, saturation, brightness, "RPC");
 
   return Status::OK();
 }
 
+void RGBWLight::ResetAutoOff() {
+  auto_off_timer_.Reset(cfg_->auto_off_delay * 1e3, 0);
```
rojer, line 384: i can read scientific notation but please... 1000 is much more readable :)

src/shelly_hap_rgb.cpp
```diff
@@ -330,42 +374,59 @@ Status RGBWLight::SetState(const std::string &state_json) {
...
     return;
   }
-  cfg_->state = false;
-  SetOutputState("auto_off");
+  TurnOff("auto_off");
+}
+
+void RGBWLight::TransitionTimerCB() {
+  int64_t elapsed = mgos_uptime_micros() - transition_start_;
+  int64_t duration = cfg_->transition_time * 1e3;
```
rojer, line 410: 1e3 -> 1000

src/shelly_hap_rgb.cpp
```diff
@@ -330,42 +374,59 @@ Status RGBWLight::SetState(const std::string &state_json) {
              "{state: %B, brightness: %d, hue: %d, saturation: %d}", &state,
              &brightness, &hue, &saturation);
```
rojer, line 376: state, brightness, hue and saturation may be used uninitialized.
remember to always initialize your variables before scanning them from json,
then validate that required keys were passed. SetState methods of other components.

rojer (changes requested):
added comments

src/shelly_hap_rgb.cpp
rojer: ok, so when you set it to on it goes to the configured brightness. makes sense.

jobl1978: > cfg_->state is a bool, it should be used as bool.
> 
> come to think of it - why do we even need separate state when we have brightness?

cfg_->state is declared as an int currently, that's the reason why I did that. Casting int's implicit to bool with !var is an ugly thing from the C century. More modern languages does even not allow that. The other thing is why is it a int currently? If it would be a real bool than this != 0 cast would not be needed.

As Timo said, on/off and brightness are different things in HK, you can change the brightness without turning on the light. if you turn on the light after, it will ramp to the previously set brightness.

timoschilling: brightness and state (aka On char) are different thinks in hk, a lamp can be off and brightness 100% at the same time, but led state should be off then.

timoschilling: Why do you think state is a int and not a Boolean?
```
void mgos_config_sw_set_defaults(struct mgos_config_sw *cfg) {
  cfg->name = NULL;
  cfg->enable = true;
  cfg->out_inverted = false;
  cfg->in_mode = 1;
  cfg->in_inverted = false;
  cfg->state = false;
  cfg->svc_type = 0;
  cfg->valve_type = 1;
  cfg->initial_state = 3;
  cfg->auto_off = false;
  cfg->auto_off_delay = 0.0;
  cfg->state_led_en = -1;
  cfg->persist_state = false;
}
```
Form `gen/mgos_config.c`, or did I miss something?

rojer: @jobl1978 
> cfg_->state is declared as an int currently

```
- ["lb.state", "b", false, {title: "State of the light"}]
```

looks like a bool to me. am i looking at the wrong struct?

rojer: oh, i see. the underlying value for config bools is an int, historically. ok, never mind.

timoschilling: Looks like a Boole is a int under the hood.
```
/* sw type struct mgos_config_sw */
struct mgos_config_sw {
  const char * name;
  int enable;
  int out_inverted;
  int in_mode;
  int in_inverted;
  int state;
  int svc_type;
  int valve_type;
  int initial_state;
  int auto_off;
  double auto_off_delay;
  int state_led_en;
  int persist_state;
};
```
from `gen/mgos_config.h`

rojer: yeah, yeah, you're right, i completely forgot about it. originally i made it so i wouldn't need to deal with sizeof(bool) being potentially different, but it seems pretty stable at 1 byte. however, i don't think we'll win any space if we switch, as fields will be 4 byte aligned anyway...

anyhow, i usually trat bool fields a bool in spirit, if not in the letter :)
i.e., i would use cfg_->state in logical context as if it were a bool, without these extra IsOn / IsOff helpers. the less code, the better.

src/shelly_hap_rgb.cpp
jobl1978: Ok I will change it to 1000, but would you say this also for 1000000? I saw it several times :-)

rojer: i didn't give it much thought but in all cases new code should be kept consistent with existing code, unless old code is an obvious antipattern (in which case it should be refactored).
this one is not a gross antipattern, more a matter of taste. i don't see pressing need for changing it, i don't find 1e6 vastly more readable, maybe because while i'm familiar with it, i'm not using scientific notation every day and it takes a moment to realize it's a number, whereas "1000000" is burned into my brain already.

src/shelly_hap_rgb.hpp
jobl1978: The reason why it is used in that way is that the HSV struct deals with ranges between 0.0f - 1.0f for every value for internally calculations. The HSV values which will come in or out has the range 0-360, 0-100, 0-100. The indention was to prevent mixing different ranges with same variables. Maybe we should implement conversion methods to the struct to deal with different ranges with the same HSV struct?

timoschilling: Whats about changing the hsv to rgbw calc to always deal with the hk ranges?

src/shelly_hap_rgb.hpp
rojer: yes, SaveState is invoked periodically from the main status timer, it's only expected to actually save state when there are changes.

src/shelly_hap_rgb.hpp
jobl1978: > Whats about changing the hsv to rgbw calc to always deal with the hk ranges?

I think this would be a clean way to fix it

src/shelly_hap_rgb.hpp
jobl1978: > scalar class members should have defaults to ensure they are always initialized.

you are right, I forgot that

jobl1978: I removed the HSV struct because I did not see any advantages with it anymore. The HSV base is now the cfg_ struct again as before.

src/shelly_hap_rgb.hpp
jobl1978: I removed the HSV struct because I did not see any advantages with it anymore. The HSV base is now the cfg_ struct again as before.

src/shelly_hap_rgb.cpp
```diff
@@ -212,43 +215,76 @@ void RGBWLight::HSVtoRGBW(const HSV &hsv, RGBW &rgbw) const {
   }
 }
 
-void RGBWLight::SetOutputState(const char *source) {
-  LOG(LL_INFO,
-      ("state: %s, brightness: %i, hue: %i, saturation: %i", OnOff(cfg_->state),
-       cfg_->brightness, cfg_->hue, cfg_->saturation));
+void RGBWLight::UpdateOnOff(bool on, const std::string &source) {
+  if (cfg_->state != on) {
```
rojer, line 219: pattern: reduce indentation to improve readability. deal with rare / error condition first and bail , otherwise continue with the main flow at the same level on indent. makes code easier to follow.

```
void RGBWLight::UpdateOnOff(bool on, const std::string &source) {
  if (cfg_->state == on) return;
  // deal with other case
}
```

src/shelly_hap_rgb.cpp
```diff
@@ -212,43 +215,76 @@ void RGBWLight::HSVtoRGBW(const HSV &hsv, RGBW &rgbw) const {
...
-    HSV hsv{.h = cfg_->hue / 360.0f,
-            .s = cfg_->saturation / 100.0f,
-            .v = cfg_->brightness / 100.0f};
+    StartTransition();
+  }
+}
 
-    HSVtoRGBW(hsv, rgbw);
+void RGBWLight::SetHue(int hue, const std::string &source) {
+  if (cfg_->hue != hue) {
```
rojer, line 235: same here

src/shelly_hap_rgb.cpp
```diff
@@ -212,43 +215,76 @@ void RGBWLight::HSVtoRGBW(const HSV &hsv, RGBW &rgbw) const {
...
+    state_notify_chars_[2]->RaiseEvent();
+    StartTransition();
   }
+}
 
-  out_r_->SetStatePWM(rgbw.r, source);
-  out_g_->SetStatePWM(rgbw.g, source);
-  out_b_->SetStatePWM(rgbw.b, source);
+void RGBWLight::SetSaturation(int saturation, const std::string &source) {
+  if (cfg_->saturation != saturation) {
```
rojer, line 246: same here

src/shelly_hap_rgb.cpp
```diff
@@ -212,43 +215,76 @@ void RGBWLight::HSVtoRGBW(const HSV &hsv, RGBW &rgbw) const {
...
+    dirty_ = true;
+    state_notify_chars_[3]->RaiseEvent();
+    StartTransition();
+  }
+}
 
-  if (out_w_ != nullptr) {
-    out_w_->SetStatePWM(rgbw.w, source);
+void RGBWLight::SetBrightness(int brightness, const std::string &source) {
+  if (cfg_->brightness != brightness) {
```
rojer, line 257: and here

src/shelly_hap_rgb.cpp
```diff
@@ -324,48 +361,91 @@ void RGBWLight::SaveState() {
...
   }
 
-  if (dirty_) {
-    SetOutputState("RPC");
+  if (brightness < 0 || brightness > 100) {
+    return mgos::Errorf(STATUS_INVALID_ARGUMENT,
+                        "invalid brightness: %d (only 0-100)", brightness);
   }
 
+  UpdateOnOff(state, "RPC");
```
rojer, line 386: if `state` is not provided, it goes to off (the default).
but what if i only want to change brightness?
if you look at other SetState implementations, they only change things that were requested. meaning, if there was no `state` key in the JSON, state should remain the same. it's a bit tricky with bools, you need to use an int8_t as the variable to use "not set" as default - see e.g. "in_inverted" in SetConfig method.
same applies to hue, saturation and brightness - if not provided, they should remain the same. you can use negative defaults for that.

rojer (changes requested):

rojer:
very good, thanks!

PR #616, opened 2021-04-08
add flash note for RGBW2

timoschilling:


rojer:
Will merge when pushing final release, the url doesn't actually work yet.

PR #630, opened 2021-04-14
Fix initial state of RGBW2

jobl1978:
RGBW2 wasn't able to recover the light (off -> on)  if it was in momentary mode with initial state last

### DIFF
--- a/src/shelly_hap_rgb.cpp
+++ b/src/shelly_hap_rgb.cpp
@@ -83,7 +83,7 @@ Status RGBWLight::Init() {
   if (IsSoftReboot()) should_restore = true;
 
   if (should_restore) {
-    UpdateOnOff(IsOn(), "init");
+    UpdateOnOff(IsOn(), "init", true);
   } else {
     switch (static_cast<InitialState>(cfg_->initial_state)) {
       case InitialState::kOff:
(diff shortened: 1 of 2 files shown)


timoschilling:
Can you revert the code order change please, the code isn't comparable in this way or split it into two commits.

jobl1978:
> Can you revert the code order change please, the code isn't comparable in this way or split it into two commits.

Pull request is now splitted into 2 commits

src/shelly_hap_rgb.cpp
```diff
@@ -145,6 +116,37 @@ Status RGBWLight::Init() {
...
+    handler_id_ =
+        in_->AddHandler(std::bind(&RGBWLight::InputEventHandler, this, _1, _2));
+    in_->SetInvert(cfg_->in_inverted);
+  }
+
+  bool should_restore = (cfg_->initial_state == (int) InitialState::kLast);
+  if (IsSoftReboot()) should_restore = true;
+
+  if (should_restore) {
+    UpdateOnOff(IsOn(), "init", true);
```
timoschilling, line 129:
```suggestion
    UpdateOnOff(IsOn(), "init", true /* force */);
```

rojer:
LGTM, waiting for Timo

timoschilling:
I have no test setup for this, but LGTM

Issue #639, opened 2021-04-21
roller shutter with legrand valena life push button


zsolt-dev:
I bought this roller shutter buttons from Legrand, since IMO they have the best looking design:
![image](https://user-images.githubusercontent.com/18726534/115614071-7cd1b880-a2ed-11eb-9931-a8d7d3f5b880.png)

They have 3 buttons: up, down and stop (the button in the middle).
It connect to just two cables (up&down).

How does the stop button works:
1. If you press the down button, then the stop button, it sends a signal to the UP wire
2. if you press up and then the stop button, it sends the signal to the DOWN wire.

So it basically remembers the last choice. It is probably mechanical, since I do not think there is any electronics inside.

So I basically need the shelly to stop the rollers when I press the opposite direction button, instead of the same direction button.

The question is:
Can we have one more mode, called "seperate - momentary (reverse stop)" ?

Thank you

link to the roller button:
https://www.legrand.com/ecatalogue/752230-roller-blind-push-button-valena.html

louis49:
I've the same problem with roller shutter buttons from Schneider Electric that have a better looking design   https://www.se.com/ww/en/product/S520208/odace%2C-rocker-switch%2C-3-positions%2C-6-ax%2C-white/

![31f89fcbe1c8be2256eeacd8db924dc9](https://user-images.githubusercontent.com/357595/115666764-432c9c00-a345-11eb-8c3a-4ab0839252c4.png)

zsolt-dev:
I did not see those anywhere... They look great. 

The existence of these validate the need for the reverse stop mode.

jobl1978:
I think having a third button (stop) for window covering is a design fail. There is no need to have a dedicated button for stop. Normally window cover switches work in that way if you press down/up window cover moves down/up. If window cover is already in move an additional press to up/down stops it. The actual Shelly HK implementation only stops if you press in the same direction as the motor currently moves.

This one is window covering switch from Berker
![window-switch](https://user-images.githubusercontent.com/33568257/115761674-e52ea180-a3a2-11eb-8b30-8f2d63a5cb63.jpg)

zsolt-dev:
> The actual Shelly HK implementation only stops if you press in the same direction as the motor currently moves. In my opinion the HK firmware should be changed that pressing a button should always stop a already moving window cover before changing the move direction.

I agree, this should be the solution.

jobl1978:
[...]

PR #643, opened 2021-04-22
Pressing up or down will always stop a moving window covering

jobl1978:
Currently only a press to the same direction will stop a moving covering.

I changed the behavior in that way that pressing any direction will stop the covering if it is moving. This is the standard behavior of most covering controllers.

Should be linked to the following issue
https://github.com/mongoose-os-apps/shelly-homekit/issues/639


rojer:
agree, thanks!

PR #649, opened 2021-04-26
Force initial state switching for ShellyRGBW2

### LINKED ISSUE #645: shelly RGB (2.9.0-alpha3)- initial state:ON - not working after the power is off and back on
First, thank you for now supporting Shelly RGB! I'm testing it and it works great. 

One thing: after you turn off the electricity and then turn it back on, shelly "forgets" the last setting (at least brightness is reset to 0%). To test it - set initial state: on, select color /brightness of your choice. then disconnect it from power. When connected again, it won't turn on (at least brightness will be set to 0%). Hope this is useful.

jobl1978:
Internal caching prevented correct switching of initial state (on/off/input).
fixes #645


timoschilling:
@rojer can you make a new alpha after merge please

PR #650, opened 2021-04-26
Rgbw light bulb refactoring

timoschilling:
requires https://github.com/mongoose-os-libs/homekit-adk/pull/5

This PR contains two steps to unify all light bulb shelly devices, which will all nearly similar to the rgbw.

I'm currently working on the support for 4xWhite on the rgbw2 and support for color bulb, duo and vintage. That brings me to the point that we need to decide to write 4 different hap classes for all of devices or we will make a configurable hap class which can be used for or all of them. I would prefer this last one.

What this PR does:
1. it renames `RGBWLight` class to a more generic `LightBulb` which also will match the related / used hap accessory type
2. it changes the `state_notify_chars_` var from a vector to a struct to not depend on the position of each chars, so we can call them by name instead of a "magic number"


src/shelly_hap_light_bulb.hpp
```diff
@@ -43,6 +43,13 @@ class RGBWLight : public Component, public mgos::hap::Service {
     float w;
   };
 
+  struct NotificationCharacteristics {
```
rojer, line 46: why do we need a struct? are they ever used as a group? i think they are more or less independent.

timoschilling: No explicit need, just use them to group, I prefer this over having 5 (1 will added later for color temp) variables

timoschilling: removed

PR #690, opened 2021-05-25
support for eve power chars

### LINKED ISSUE #26: Shelly Powermeter consumption reading.
Hello, is a development planned to have the consumption data read by the Plug-S as on the Shelly Cloud app?

Thank you.

timoschilling:
closes #26

This characteristics aren't supported by the Home App, they can be view with the Eve or Home+ App.

This PR explicitly don't contains support for the eve power history char. That is out of scope for this PR, because it requires general change like storing the power consumption first.


src/shelly_switch.cpp
```diff
@@ -25,6 +25,14 @@
 
 namespace shelly {
 
+const HAPUUID kHAPCharacteristic_EveConsumption = {
```
rojer, line 28: please add comments on what these are.

src/shelly_switch.cpp
```diff
@@ -25,6 +25,14 @@
...
+    0x52, 0x9F, 0xA2, 0x05, 0x26, 0x9C, 0x27, 0x8F,
+    0xFF, 0x48, 0x9E, 0x07, 0x0D, 0xF1, 0x63, 0xE8};
```
rojer, line 30: please move closing brace to the next line, ditto below.

src/shelly_switch.cpp
rojer: and make both static.

src/shelly_switch.cpp
```diff
@@ -316,4 +325,36 @@ void ShellySwitch::InputEventHandler(Input::Event ev, bool state) {
...
+      (*iid)++, &kHAPCharacteristic_EveTotalConsumption, 0, 65535, 1,
+      [this](HAPAccessoryServerRef *,
+             const HAPUInt16CharacteristicReadRequest *, uint16_t *value) {
+        *value = out_pm_->GetEnergyWH().ValueOrDie() / 1000.0f;
+        return kHAPError_None;
+      },
+      true /* supports_notification */, nullptr, "eve-total-power-consumption");
+  AddChar(total_power_char);
+
+  power_timer_.Reset(1000, MGOS_TIMER_REPEAT);
```
rojer, line 352: 1 second seems too often, will cause a lot of traffic.
please change to, like 10 seconds and add some fuzz, e.g. +/- 500 ms (use [mgos_rand_range](mgos_rand_range)).

src/shelly_switch.hpp
```diff
@@ -74,6 +74,12 @@ class ShellySwitch : public Component, public mgos::hap::Service {
   bool dirty_ = false;
 
   ShellySwitch(const ShellySwitch &other) = delete;
+
+  void AddPowerMeter(uint16_t *iid);
+  void PowerMeterTimerCB();
+  mgos::Timer power_timer_;
+  mgos::hap::Characteristic *power_char;
```
rojer, line 81: please add underscore at the end, ditto below.

rojer (changes requested):

src/shelly_switch.cpp
timoschilling: That’s what `make format` does

rojer: leave a comma at the end and it won't

src/shelly_switch.cpp
timoschilling: I found already 1 sec long. What do you think about keeping 1 sec but only firing an event if value is changed?

src/shelly_switch.cpp
timoschilling: What kind of comment do you mean? For me the name says enough, please make a suggestion.

About making them static, I don't know how, `static const` wouldn't compile.

Note: https://github.com/mongoose-os-apps/shelly-homekit/blob/master/src/shelly_hap_doorbell.cpp#L25 uses the same notation as here, no static, no comment.

rojer: magic numbers need comments, there are a lot of magic numbers here, we need a link to where they come from, to give context.
the doorbell thing could use a comment too - saying standalone doorbell service is deprecated, etc.

src/shelly_switch.cpp
rojer: firing if consumption changes SGTM, thats' exactly how it's supposed to work.

timoschilling:
@rojer any blockers left?

src/shelly_switch.hpp
```diff
@@ -74,6 +74,14 @@ class ShellySwitch : public Component, public mgos::hap::Service {
   bool dirty_ = false;
 
   ShellySwitch(const ShellySwitch &other) = delete;
+
+  void AddPowerMeter(uint16_t *iid);
+  void PowerMeterTimerCB();
+  mgos::Timer power_timer_;
+  mgos::hap::Characteristic *power_char_;
+  mgos::hap::Characteristic *total_power_char__;
```
rojer, line 82: one underscore is sufficient.
also please initialize with nullptr

src/shelly_switch.cpp
```diff
@@ -316,4 +334,46 @@ void ShellySwitch::InputEventHandler(Input::Event ev, bool state) {
...
+        return kHAPError_None;
+      },
+      true /* supports_notification */, nullptr, "eve-total-power-consumption");
+  AddChar(total_power_char__);
+
+  power_timer_.Reset(1000, MGOS_TIMER_REPEAT);
+}
+
+void ShellySwitch::PowerMeterTimerCB() {
+  float current_power = out_pm_->GetPowerW().ValueOrDie();
```
rojer, line 365: careful there, you don't want to die just because power meter had a hiccup.
always check if returned value is ok and take evasive action.
no dying please :)

rojer (changes requested):
can you please check if adding these characteritics causes devices in home app to be re-enumerated and lose their customizations (like assigned name)? i don't think it will, but want to be sure.

src/shelly_switch.hpp
timoschilling: oh ©️ 🍝

src/shelly_switch.cpp
rojer: wdym? it's never ok to call ValueOrDie without checking for .ok() first.
i.e. you must always make sure the program won't actually die, it's never of for firmware to keel over and abort.
"OrDie" in the name is there to remind you of that.

src/shelly_switch.cpp
```diff
@@ -316,4 +334,52 @@ void ShellySwitch::InputEventHandler(Input::Event ev, bool state) {
...
+void ShellySwitch::AddPowerMeter(uint16_t *iid) {
+  if (out_pm_ == nullptr) return;
+
+  // Power
+  power_char_ = new mgos::hap::UInt16Characteristic(
+      (*iid)++, &kHAPCharacteristic_EveConsumption, 0, 65535, 1,
+      [this](HAPAccessoryServerRef *,
+             const HAPUInt16CharacteristicReadRequest *, uint16_t *value) {
+        auto power = out_pm_->GetPowerW();
+        if (!power.ok())
```
rojer, line 346: please put braces around the body of the if statement, since it's multi-line.
ditto below.

rojer (changes requested):

src/shelly_switch.cpp
timoschilling: It's not multiline, it's a guard clause which returns an error if not ok, if ok value will be set and "no error" will be returned.

rojer: well, it's on two lines, that's multiline in my book :)
omitting blaces is only ok if entire statement fits on one line
```
if (!energy.ok()) return kHAPError_Busy;
```

here we need braces to remove any possibility of control flow being wrong if statements are added later.
```
if (!energy.ok()) {
  return kHAPError_Busy;  // Operation failed temporarily, retry later.
}
```

timoschilling: Ok know what you mean. What do you think, does we need the comment? Without comment we can use a real oneliner.

rojer:
very good, thanks!

PR #699, opened 2021-06-07
use minified qrcode.js

timoschilling:
the minified version from GitHub is broken, so I minified the regular version by my own.


rojer:
thank you!

rojer:
3K shaved off index.html.gz, yay!

PR #706, opened 2021-06-15
remove redundant w channel nulling

timoschilling:
the unused w channel is already nulled here https://github.com/mongoose-os-apps/shelly-homekit/blob/master/src/ShellyRGBW2/shelly_init.cpp#L57


rojer:
it's not immediately obvious to me that SetStatePWM on an output zeroes the white channel value in this struct (i don't think it does, actually).
i think ensuring that rgbw.w is set to zero if there is no white output channel is good.

PR #718, opened 2021-07-02
added support for Prometheus for output, powermanagement and temperature

nikwest:
I added Prometheus support to some of the shelly properties. This allows me to monitor e.g. power usage with Grafana.
This is just a quick shot, I'm not sure if you could do the callbacks as instance member functions of the C++ class. That would probably be nicer, but I'm not very familiar with C++. 
If the prometheus library is not added the specific code should not be compiled. At least that was my understanding from the prometheus-metrics library.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 MAKEFLAGS += --warn-undefined-variables
 
-.PHONY: build check-format format release upload Shelly1 Shelly1L Shelly1PM Shelly25 Shelly2 ShellyI3 ShellyPlug ShellyPlugS ShellyRGBW2 Custom2
+.PHONY: build check-format format release upload Shelly1 Shelly1L Shelly1PM Shelly25 Shelly2 ShellyI3 ShellyPlug ShellyPlugS ShellyRGBW2 Custom2 Custom16
 
 MOS ?= mos
 # Build locally by default if Docker is available.
@@ -24,7 +24,7 @@ ifneq "$(VERBOSE)$(V)" "00"
   MOS_BUILD_FLAGS_FINAL += --verbose
 endif
 
-build: Shelly1 Shelly1L Shelly1PM Shelly2 Shelly25 ShellyI3 ShellyPlug ShellyPlugS ShellyRGBW2 ShellyU ShellyU25 Custom2
+build: Shelly1 Shelly1L Shelly1PM Shelly2 Shelly25 ShellyI3 ShellyPlug ShellyPlugS ShellyRGBW2 ShellyU ShellyU25 Custom2 Custom16
 
 release:
 	@[ -z "$(wildcard fs/conf*.json fs/kvs.json)" ] || { echo; echo "XXX No configs in release builds allowed"; echo; exit 1; }
@@ -68,6 +68,10 @@ ShellyU25: build-ShellyU25
 	@true
 
 Custom2: build-Custom2
+	@true
+
+Custom16: PLATFORM=esp32
+Custom16: build-Custom16
 	@true
 
 fs/index.html.gz: $(wildcard fs_src/*) Makefile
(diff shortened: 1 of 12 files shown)


timoschilling:
IMO supporting other "data places" then HomeKit is out of scope for this project. If someone wants to collect data in a 3rd party system, he should collect the data by its own (via reading the api or log). If this will be merged, someone else will request support for the next logging system and some day we will end up supporting dozens of systems.

nikwest:
I get your point, however Prometheus/Grafana combo is a pretty well established stack for building nice and powerful dashboards, it's not about logging.
Why would anybody buy a Shelly 1PM or similar power measuring device to just use it with HomeKit alone, which is not even able to display any of the power measuring information?

timoschilling:
> Why would anybody buy a Shelly 1PM or similar power measuring device to just use it with HomeKit alone, which is not even able to display any of the power measuring information?

Isn't right, since https://github.com/mongoose-os-apps/shelly-homekit/pull/690 we support power measuring information in HomeKit. The Home App doesn't support it but multiple 3rd party apps does it. That's how HomeKit is designed, it's just a distributed key value store and the data protocol has some predefinitions but is also open for 3rd party extensions.

I think in the end logging, dashboard and what ever are the same, all are data places which we need to support. We had already requests for supporting homebridge, iobroker and co in the past, for them counting the same arguments, we can't support any 3rd party system. If you wan't to get the shelly data into a 3rd party system build a plugin for that system, because that system is build for data handling or build a broker in between.

Some side question how do you monitor other HomeKit devices which didn't have a open source firmware?

IMO HomeKit is already the unified data place, if you want to get data of a HomeKit device read it from HomeKit so you can monitor all your HomeKit devices and not only the shelly's.

But all that is just my opinion, in the end it's up to @rojer to make a decision.

nikwest:
As I said I understand your position and I don't want to start a discussion about it. I personally have the strong opinion that there is never going to be something like a unified data place. I've been in IT for too long and never saw it happen. So the key question is, if the project wants to be a one trick pony or be as broad as possible. It's really a strategic question. 
The prometheus integration is light weight, all is done in mongoose anyways, so it's just about calling the functionality at the right places. I also like it, because it gives you a human readable, console and scripting friendly output. So even if you are not using Prometheus itself the /metrics URL is very helpful.

rojer:
homekit support is the primary goal of this project.
this is and will be our favorite trick, but it need not be the one and only.
as explained in the aws and mqtt support request, i am potentially open to adding more integrations provided they don't interfere with HK (i.e. don't substantially change the architecture and make things more difficult for HK support) and don't use much space, which is in short supply.

prometheus support makes sense to me overall.

the code change seems lightweight but what are the space requirements? can you post code size (.bin file in the build dir) before and after this change? if it's more than a few kb, we're not doing this (yes, we are very tight on space).

if the size is ok, then we can iterate on the implementation (i have a few comments).

nikwest:
I did a build for the Shelly25 and here is the outcome:


**current master:**
`725520  objs/switch25.bin
`
**with Prometheus:**
` 729616  objs/switch25.bin
`
So it is about 4.1k difference.

rojer:
our max is 716K, i.e. 733184 so this eats half of our remaining budget. i don't think it's worth it, at this point.
we do have a plan for when we run out of space eventually but i'd like to postpone it as much as possible.
i would be ok with getting it in but compile-time disabled by default, so you can have custom firmware with it if you like.
we can consider enabling it by default when we reduce the FS size and get another 128K of room.

nikwest:
Yes, I think it makes sense to just add it to the code but have it disabled by default. From my side I would try to still improve the implementation a bit. I have used it with a single device for some weeks now and it works nicely, but if you have more devices I think it's better to add some information like name of the device and switch to be useable within queries. Right now it's only the IP address and that's not very comfortable to use.

timoschilling:
@nikwest Why you’re not going this way? https://github.com/mongoose-os-apps/shelly-homekit/issues/722

timoschilling:
As this PR looks more like a private fork now, I would like to close it. @nikwest feel free to open a new PR with a clean version and Rojers suggestions applied.

PR #731, opened 2021-07-12
Add set position input to shutter forms

ChaosGramer:
Hello everyone, 

this PR contains a field on the form of the shutter which allows the user to set the target position of the shutter. 

Looking forward to hear from you. Regards, Jan

### DIFF
--- a/fs_src/index.html
+++ b/fs_src/index.html
@@ -540,6 +540,13 @@
           <label><span id="close_spinner"></span>Close</label>
         </button>
       </div>
+      <div class="form-control" id="set_tgt_pos_form">
+        <label>Set to Postion</label>
+        <input type="number" id="set_tgt_pos_input">
+        <button id="set_tgt_pos_button">
+          <label><span id="open_spinner"></span>Set</label>
+        </button>
+      </div>
       <div class="form-control">
         <label>Calibration:</label>
         <span id="cal"></span>
--- a/fs_src/script.js
+++ b/fs_src/script.js
@@ -489,6 +489,9 @@ function findOrAddContainer(cd) {
       el(c, "close_btn").onclick = function () {
         setComponentState(c, {tgt_pos: 0}, el(c, "close_spinner"));
       };
+      el(c, "set_tgt_pos_button").onclick = function () {
+        setComponentState(c, {tgt_pos: el(c, "set_tgt_pos_input").value}, el(c, "close_spinner"));
+      }
       el(c, "save_btn").onclick = function () {
         wcSetConfig(c, null, el(c, "save_spinner"))
       };
--- a/fs_src/style.css
+++ b/fs_src/style.css
@@ -284,3 +284,11 @@ input:checked + .slider:before {
 #hap_setup_code {
   font-weight: bold;
 }
+
+#set_tgt_pos_form {
+  display: flex;
+}
+
+#set_tgt_pos_input {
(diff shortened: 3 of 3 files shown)


fs_src/index.html
```diff
@@ -540,6 +540,13 @@ <h1 id="head">Window</h1>
           <label><span id="close_spinner"></span>Close</label>
         </button>
       </div>
+      <form class="form-control" id="set_tgt_pos_form">
+        <label>Set to Postion</label>
+        <input type="number" id="set_tgt_pos_input">
+        <button id="set_tgt_pos_button" type="submit">
```
timoschilling, line 546: Remove the type=submit

fs_src/script.js
```diff
@@ -489,6 +489,10 @@ function findOrAddContainer(cd) {
       el(c, "close_btn").onclick = function () {
         setComponentState(c, {tgt_pos: 0}, el(c, "close_spinner"));
       };
+      el(c, "set_tgt_pos_form").onsubmit = function (e) {
```
timoschilling, line 492: Change onsubmit to onclick

fs_src/script.js
```diff
@@ -489,6 +489,10 @@ function findOrAddContainer(cd) {
...
+        setComponentState(c, {tgt_pos: el(c, "set_tgt_pos_input").value}, el(c, "close_spinner"));
+        e.preventDefault();
```
timoschilling, line 494: Can be removed

timoschilling (changes requested):
Not tested but I expect that all requests together should work

fs_src/index.html
ChaosGramer: This way the enter key would not work that is why I choose to go for form + submit. Is this breaking anything?

timoschilling: I see what you mean, but then this field will behave different then any other, so we should not add this feature to just one of them

ChaosGramer: okay I see, I can adjust the other fields too, this will be straight forward.

timoschilling: Maybe in a different PR

ChaosGramer: okay I adjusted the code, please review again.

fs_src/script.js
ChaosGramer: done

fs_src/index.html
timoschilling: You missed to remove the type attr

ChaosGramer: good catch

ChaosGramer:
all comments solved I guess?

timoschilling:
LGTM, but I’m on vacation and not able to test it

@rojer can you take a look

rojer:
looks ok, thanks!

rojer:
built and deployed the new fw, a couple comments:
 * please make the input smaller. it will only ever allow a 3 digit number, it doesn't need to be that big.
 * please change the label to "Position:" and center vertically

rojer:
@ChaosGramer will you be able to send a followup or should I do it myself?

ChaosGramer:
I tried already to fix it locally gimme the upcoming weekend looks like flex. box is not desirable for that purpose. 

Yeah I will send the follow up soon.

rojer:
i needed to push 2.9.4 and reverted your commit temporarily, please resubmit with the changes.

Issue #939, opened 2022-02-18
POC: Bulb reset


timoschilling:
@rojer my idea for a bulb reset mechanism is to add a `shelly.reboot_counter = 0` to the bulbs config.
```cpp
void InitApp() {
  if (mgos_sys_config-> shelly_reboot_counter >= 3) {
    DoReset();
  }
  mgos_sys_config-> shelly_reboot_counter++;
  mgos_sys_config_save(&mgos_sys_config, false, nullptr)
  mgos_set_timer(10000, 0, ResetRebootCounter, NULL);
```

What do you think about it?
Should we only add it to bulbs or to all shelly's, I see a benefit having it on all shelly's, sometime use have shelly's without switches connected behind a wall and can't reset them.

MilanBacaCz:
It is greate idea. I have many shellys. They work well, but sometimes they are unresponsive and i have to Power off them. Then they work well again. I tried reconnect them in the router, but it did not help. Only Power off and on make them work again. So reset timer is good idea.

MagOO33:
Great!
same kind of idea, could it be possible to store the last command in order to show the good state if unwanted reboot appears because of power spikes.
The case is shelly 2.5 controlling roller shutters and launching a up or down command and rebooting because of inductive spike.
When it reboots, it shows the wrong state because command was done by relay while rebooting.
This could avoid a calibration to make roller shutter rework

timoschilling:
closed because PR is open

PR #941, opened 2022-02-20
Boot reset

### LINKED ISSUE #918: Reset mechanism for bulbs
Bulbs have no buttons or external inputs. We need some sort of a reset mechanism for them.

timoschilling:
This is primary for the bulb as they don't have any reset button, but I and some other users (see https://github.com/mongoose-os-apps/shelly-homekit/issues/939) found it useful to have it on all devices.

closes #918

### DIFF
--- a/mos.yml
+++ b/mos.yml
@@ -37,6 +37,7 @@ config_schema:
   - ["shelly.legacy_hap_layout", "b", false, {title: "Use legacy accessory layout instead of a bridged accessory"}]
   - ["shelly.overheat_on", "i", 100, {title: "Overheat protection mode kicks in at or above this temperature"}]
   - ["shelly.overheat_off", "i", 90, {title: "Overheat protection mode turns off when the temperature is back below this threshold"}]
+  - ["shelly.reboot_counter", "i", 0, {title: "Counter of boot tries with a uptime of less then 10 sec."}]
 
   - ["ts", "o", {title: "Temperature Sensor settings", abstract: true}]
   - ["ts.name", "s", "", {title: "Name of the sensor"}]
--- a/src/shelly_main.cpp
+++ b/src/shelly_main.cpp
@@ -835,7 +835,24 @@ static void HTTPHandler(struct mg_connection *nc, int ev, void *ev_data,
   (void) user_data;
 }
 
+void ResetRebootCounter(void *arg UNUSED_ARG) {
+  LOG(LL_INFO, ("=== ResetRebootCounter"));
+  mgos_sys_config_set_shelly_reboot_counter(0);
+  mgos_sys_config_save(&mgos_sys_config, false /* try_once */, nullptr);
+}
+
 void InitApp() {
+  int reboot_counter = mgos_sys_config_get_shelly_reboot_counter() + 1;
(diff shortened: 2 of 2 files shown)


rojer:
haven't looked at the code yet, but a general comment: we should match the stock approach, whatever it is.

timoschilling:
This is from the Vintage manual, there other bulbs have similar texts:
> FACTORY RESET
> You can return your Shelly Vintage to its factory settings by switching the switch/button to which the bulb is connected 5 times. You need to wait 3 seconds after every switch off and 1 second after every switch on. If your bulb is successfully reset it will blink few times and then turn the light on.

The switches didn't have a text like that, but I still thing it's useful and they should have the functionality.

rojer:
ok, this seems reasonable. two things:
1) the code belongs in libreset
2) we need this part:
> If your bulb is successfully reset it will blink few times and then turn the light on.

to provide this feedback to the user that reset has been performed.
bulbs/rgb components should implement this "blink a few times at max brighntess" mechanism, which can also be used for identification.

timoschilling:
> the code belongs in libreset

I think moving it over to libreset is out of scope for this PR, it has a dependency to DoReset, which has a dependency to CheckLED. So simply moving it over to libreset didn't work.

rojer:
then we should broaden the scope of the PR. it's a piece reset-related code that belongs in libreset. if other things need to be refactored for taht to happen - so be it, make it so.

timoschilling:
Sorry I'm out for that. I have no idea how to solve it. This PR add's a functionality you requested and I will provide it. My code depence on DoReset and not on libreset, so I placed it where DoReset is defined. If you want that refactoring merge this PR and do it by your own.

timoschilling:
Or make you the first step and move DoReset into libreset and then I will move my code there as well.

rojer:
hi Timo. forst of all - apologies, i realize my comment may have sounded a bit offensive.
this was unintentional, i am simply too distraught with all the events around ukraine, i can barely concentrate (i have relatives on both sides, it's a horrible tragedy for me personally). believe me, i appreciate your work on the project.
my point was: often in order to do something right some additional work is required. i personally never shy away from such work and encourage others to do things right rather than quickly, at all times. we have no firm deadlines, so time pressure is not a factor.
in this specific case, shelly_main.cc is long overdue for refactoring and we need to move things _out_ of it, not pile new stuff _in_. doing it in a separate PR sounds good, but it should happen before we add new stuff.
i am also fine with doing the cleanup myself, no problem. it won't happen immediately - as i said, i can barely concentrate right now, i've been off my job since thursday and will probably take more time off this week.

timoschilling:
Don't worry, it's ok, I know how it is with OSS projects and in general my mindset about adding only code wich is fine and better doing a refactoring now then never!
My problem is C/C++ isn't my world, it's more a try and error, read and copy process for me, so I have no I idea how you want to structure the code. If it would only be a copy thing I would do it, but there is too much unknown concept work for me. 

I have already tried to move it over, but I have no idea how to deal with DoReset, moving DoReset to libreset feels wrong because it has application logic, moving it into an own file and including it it's the same. The same counts for CeckLED.
If you have a easy solution, I will try to implement it.

I feel with you and hope your relatives stay save and all that shit will end fast and peaceful!

PR #946, opened 2022-03-02
add support for plus 1(pm) devices with hk fw to flash tool

timoschilling:
This PR add the ability to find and update plus devices which already have the HomeKit firmware.
Every think else may follow in different PRs.

### DIFF
--- a/tools/flash-shelly.py
+++ b/tools/flash-shelly.py
@@ -85,6 +85,7 @@ optional arguments:
 import argparse
 import atexit
 import datetime
+from distutils.log import debug
 import functools
 import http.server
 import ipaddress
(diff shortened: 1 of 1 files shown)


tools/flash-shelly.py
```diff
@@ -227,7 +228,7 @@ def add_service(self, zc, service_type, device):
       logger.debug(f"[Device Scan] found device: {host}, IP address: {wifi_ip}")
       logger.trace(f"[Device Scan] info: {info}")
       logger.trace(f"[Device Scan] properties: {properties}")
-      if properties.get('arch') == 'esp8266':  # this detects if esp device.
+      if properties.get('arch') == 'esp8266' or properties.get('arch') == 'esp32':  # this detects if esp device.
```
andyblac, line 231: maybe use `if properties.get('arch') in ('esp8266', 'esp32'):`, slightly cleaner.

PR #994, opened 2022-05-21
light bulb: identify by blinking three times

markirb:
Proposal for the blinking.

Not really wholly satisfied with the manipulating of the light_bulb_controller, this goes a little bit in the direction of light effects. 
so feel free to discuss cleaner solutions :)


rojer:
hm, as you said, the surgery done on the LightBulbController is not the most elegant.
creating the effects class with its own timer and stuff... let's not go there just yet.
having thought about it a bit, here's what i think might work: refactor LightBulbController manage not just start/current/end states but an array of them, and transition between them automatically.
this way effects could be expressed by just submitting a sequence of states to the controller and have it execute them.
identify will then look like a vector of `{full_on,off,full_on,off,full_on,off,current}`

markirb:
should be easier now

rojer:
@markirb hey thanks, this looks promising! there are some minor style nits, but before cleaning them, let me ask @timoschilling - Timo, what do you think? does this look reasonable?

rojer:
@markirb thank you very much, i cleaned up the code a bit and pushed to master.
@timoschilling can i ask you to test with RGBW2 please? i only tested with vintage.

PR #997, opened 2022-05-23
WC: Respect ramp up time when calibrating close

jonnylamb:
The `wc.max_ramp_up_time_ms` setting defaults to 5000ms and gives time
for motors to ramp up and draw some real power before we determine
whether they're actually running (by comparing their power draw to the
`wc.idle_power_thr` setting).

When calibrating the window covering _opening_ the calibration logic
waits the whole ramp up time before determining whether the motor is
idle, but this logic is missing when calibrating the closing. There's
just a 1000ms check hard-coded but if the motor takes longer than that
to get going and draw sufficient power the calibration will be deemed
complete (too early).

This patch simply copies the ramp up check from the open calibration
logic to the close calibration logic, which I guess was just forgotten
in f41b01f5854f5.


rojer:
good catch, thanks!

PR #1009, opened 2022-06-13
bulbs: fix identify for primary accessories

markirb:
this needs: https://github.com/mongoose-os-libs/homekit-adk/pull/9

### DIFF
--- a/src/ShellyVintage/shelly_init.cpp
+++ b/src/ShellyVintage/shelly_init.cpp
@@ -48,18 +48,17 @@ void CreateComponents(std::vector<std::unique_ptr<Component>> *comps,
   }
 
   mgos::hap::Accessory *pri_acc = accs->front().get();
+  shelly::hap::LightBulb *light_ref = hap_light.get();
   hap_light->set_primary(true);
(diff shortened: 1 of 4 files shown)


src/ShellyVintage/shelly_init.cpp
```diff
@@ -48,18 +48,17 @@ void CreateComponents(std::vector<std::unique_ptr<Component>> *comps,
...
   pri_acc->AddService(hap_light.get());
+  pri_acc->SetIdentifyCB(
+      [light_ref](const HAPAccessoryIdentifyRequest *request UNUSED_ARG) {
+        light_ref->Identify();
+        return kHAPError_None;
+      });
 
   comps->push_back(std::move(hap_light));
-
-  SetIdentifyCB([](const HAPAccessoryIdentifyRequest *request UNUSED_ARG) {
```
rojer, line 57: we still need SetIdentifyCB though, as that's what gets called first.

rojer: yes, you're right. i forgot the way it works already... yes, this looks good.

PR #1028, opened 2022-07-06
adaptive lighting support for CCT lights

markirb:
I have been working on this for a while, as I really find this an interesting feature.
I think now it is in a state for review and testing.

currently it relies on one additional fix https://github.com/mongoose-os-libs/homekit-adk/pull/10

This adds the Homekit feature adaptive lighting for Lights supporting ColorTemperature (CCT at the moment).
When adaptive lighting is enabled, the light gets a schedule every 24 hours how to adapt the color temperature according to the current night / daytime. Afterwards the schedule is ran on the device itself see e.g. https://www.howtogeek.com/712520/how-to-use-adaptive-lighting-with-apple-homekit-lights/ for how this works.

As this is not (yet) documented in the published Apple Homekit spec, this relies heavily on the reverse engineering of the TLV done by Homebridge specifically https://github.com/homebridge/HAP-NodeJS/blob/master/src/lib/controller/AdaptiveLightingController.ts.

The only thing missing is making the schedule permanent. But I do not know if this makes sense as it uses ~1kB of configuration storage and only would make sense if we have a notion of time on the device otherwise we do not know where to continue after a downtime.

### DIFF
--- a/mos.yml
+++ b/mos.yml
@@ -144,7 +144,7 @@ libs:
   - location: ./libreset
   - location: https://github.com/mongoose-os-libs/core
   - location: https://github.com/mongoose-os-libs/file-logger
-  - location: https://github.com/mongoose-os-libs/homekit-adk
+  - location: https://github.com/markirb/homekit-adk
   - location: https://github.com/mongoose-os-libs/http-server
   - location: https://github.com/mongoose-os-libs/ota-http-server
   - location: https://github.com/mongoose-os-libs/rpc-service-config
@@ -190,6 +190,7 @@ conds:
         HAP_TLV_NO_LOG: 1            # Saves us stack
         HAP_DISABLE_ASSERTS: 1       # 32K
         HAP_DISABLE_PRECONDITIONS: 1 # 40K
+
   - when: mos.platform == "esp32"
     apply:
       config_schema:
--- a/mos.yml
+++ b/mos.yml
@@ -187,9 +187,9 @@ conds:
         # XXX: fix size calculation with and without BLE
         XHAP_ACCESSORY_SERVER_SIZE: 1680
         HAP_LOG_LEVEL: 0             # This saves 36K on esp8266.
+        HAP_TLV_NO_LOG: 1            # Saves us stack
         HAP_DISABLE_ASSERTS: 1       # 32K
         HAP_DISABLE_PRECONDITIONS: 1 # 40K
-
   - when: mos.platform == "esp32"
     apply:
       config_schema:
--- a/mos.yml
+++ b/mos.yml
@@ -144,7 +144,8 @@ libs:
   - location: ./libreset
   - location: https://github.com/mongoose-os-libs/core
   - location: https://github.com/mongoose-os-libs/file-logger
-  - location: https://github.com/markirb/homekit-adk
+  - location: https://github.com/mongoose-os-libs/homekit-adk
+    version: pull/10/head # use unmerged changes from https://github.com/mongoose-os-libs/homekit-adk/pull/10
   - location: https://github.com/mongoose-os-libs/http-server
   - location: https://github.com/mongoose-os-libs/ota-http-server
   - location: https://github.com/mongoose-os-libs/rpc-service-config
--- a/src/ShellyDuo/shelly_init.cpp
+++ b/src/ShellyDuo/shelly_init.cpp
@@ -49,6 +49,14 @@ void CreateComponents(std::vector<std::unique_ptr<Component>> *comps,
     return;
   }
 
+  // Use adaptive lightning when possible (CCT)
+  std::unique_ptr<hap::AdaptiveLighting> adaptive_light;
+  adaptive_light.reset(new hap::AdaptiveLighting(hap_light.get(), lb_cfg));
+  auto st = adaptive_light->Init();
+  if (st.ok()) {
+    hap_light->SetAdaptiveLight(std::move(adaptive_light));
+  }
+
   mgos::hap::Accessory *pri_acc = accs->front().get();
   shelly::hap::LightBulb *light_ref = hap_light.get();
   hap_light->set_primary(true);
--- a/src/ShellyRGBW2/shelly_init.cpp
+++ b/src/ShellyRGBW2/shelly_init.cpp
@@ -16,6 +16,7 @@
  */
 
 #include "shelly_cct_controller.hpp"
+#include "shelly_hap_adaptive_lighting.hpp"
 #include "shelly_hap_input.hpp"
 #include "shelly_hap_light_bulb.hpp"
 #include "shelly_input_pin.hpp"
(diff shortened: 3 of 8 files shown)


mos.yml
```diff
@@ -187,9 +187,9 @@ conds:
         # XXX: fix size calculation with and without BLE
         XHAP_ACCESSORY_SERVER_SIZE: 1680
         HAP_LOG_LEVEL: 0             # This saves 36K on esp8266.
+        HAP_TLV_NO_LOG: 1            # Saves us stack
         HAP_DISABLE_ASSERTS: 1       # 32K
         HAP_DISABLE_PRECONDITIONS: 1 # 40K
-
```
timoschilling, line 192: Should be reverted

timoschilling:
Nice, didn't noticed this PR in July.

bobziroll:
Any updates on this? I'm super excited I found this firmware for the Shelly RGBW2 and was able to get my RTF-Lighting FCOB CCT lights to work with tunable controls in HomeKit. Nothing else I tried worked as well as this.

The icing on the cake would be having adaptive lighting support as well 🤩

markirb:
Mh I use this every day since several months so I can say that for me works pretty well :)

If @rojer fixes the build pipeline by updating https://hub.docker.com/r/mgos/esp32-build/tags via https://github.com/cesanta/mongoose-os/pull/603 then the GitHub action would provide some beta you could try on your device. Or you build it yourself? :)

Any feedback welcome...

markirb:
Okay, this is the current build of commit [763cd67](https://github.com/mongoose-os-apps/shelly-homekit/pull/1028/commits/763cd6772fb58c31a30ae2a88882e3c908a674ad) for uploading via web interface
[ShellyRGBW2.zip](https://github.com/mongoose-os-apps/shelly-homekit/files/10450560/ShellyRGBW2.zip)

bobziroll:
Seems to be working! I had to remove the accessory and add it back into HomeKit but it's got the adaptive lighting option there now and seems to be working great!

markirb:
Readd? Strange... Are you on the "new architecture" of HomeKit of latest iOS update (that recently got pulled again?) might be the caching of this new architecture...

bobziroll:
I'm not sure, I don't know anything about it. How would I check?

mos.yml
```diff
@@ -144,7 +144,7 @@ libs:
   - location: ./libreset
   - location: https://github.com/mongoose-os-libs/core
   - location: https://github.com/mongoose-os-libs/file-logger
-  - location: https://github.com/mongoose-os-libs/homekit-adk
+  - location: https://github.com/markirb/homekit-adk
```
timoschilling, line 147:
```suggestion
  - location: https://github.com/mongoose-os-libs/homekit-adk
    version: pull/10/head # use unmerged changes from https://github.com/mongoose-os-libs/homekit-adk/pull/10
```

src/ShellyDuo/shelly_init.cpp
```diff
@@ -49,6 +49,14 @@ void CreateComponents(std::vector<std::unique_ptr<Component>> *comps,
...
   }
 
+  // Use adaptive lightning when possible (CCT)
+  std::unique_ptr<hap::AdaptiveLighting> adaptive_light;
+  adaptive_light.reset(new hap::AdaptiveLighting(hap_light.get(), lb_cfg));
+  auto st = adaptive_light->Init();
+  if (st.ok()) {
+    hap_light->SetAdaptiveLight(std::move(adaptive_light));
+  }
+
```
timoschilling, line 59: Could this be integrated in the `CCTController`? Because it's the only light which supports it.

markirb:
> I'm not sure, I don't know anything about it. How would I check?
> 

I guess currently it's complicated if you don't already know... so never mind.

bobziroll:
@markirb I wasn't on the new architecture at the time. I just updated yesterday, so now I am. No issues so far.

src/ShellyDuo/shelly_init.cpp
markirb: I guess it could. But on the other hand: the controller should not need to know about adaptive lights existing. And the Adaptive Lightning Protocol of Apple is generic in the way that it could support autotuning other characteristics, not limited to only color temperature, so I actually think it would make more sense this way

timoschilling: Good argument. My main point is that I don't like ot have redundant code in the device specific files. You follow the current pattern, so it's ok. That problem should be solved, but not in this PR.

markirb: agree

bobziroll:
I tried installing the firmware from `http://A.B.C.D/ota?url=http://shelly.rojer.cloud/update` last night on my 2nd RGBW2 device after I saw this feature had been merged, but adaptive lighting still didn't show up as an option on HomeKit. Did I miss something? I made sure the firmware was updated before doing it.

I just ended up just installing the zip file from @markirb on it like I did on my first RGBW2 device instead.

markirb:
pushing to master does not equal a release.

first it will go to a beta, then somewhen a release

timoschilling:
Here are some prerelease builds with this PR, for the devices which are supporting CCT.
[shelly-homekit-ShellyRGBW2.zip](https://github.com/mongoose-os-apps/shelly-homekit/files/10522565/shelly-homekit-ShellyRGBW2.zip)
[shelly-homekit-ShellyDuo.zip](https://github.com/mongoose-os-apps/shelly-homekit/files/10522566/shelly-homekit-ShellyDuo.zip)
(ping me if I missed one)

bobziroll:
> pushing to master does not equal a release. first it will go to a beta, then somewhen a release.

@markirb Ah, okay makes sense.

> Here are some prerelease builds with this PR, for the devices which are supporting CCT.

@timoschilling Thanks so much! They seem to be working great so far.

InventoryAndi:
Hi @markirb: is it there also a possibility to support this feature for Shelly Duo - RGBW? I just know that there are nanoleaf lights available which supports RGB and CCT and you can activate adaptive lighting in homekit. Not sure if it is a different technology in comparison to Shelly but I think Shelly Duo - RGBW have also cold and warm white so I think this feature could be possible?

InventoryAndi:
Hi. The idea was to enable adaptive light for the Shelly lamp RGBW (as it is available for CCT). At the moment it only works for the CCT.
I thought this might be possible because I have a nano leaf bulb RGBCW and for this bulb the adaptive light is also available. But maybe I'm wrong and the nano leaf is a RGBCCT and that's why for this build the adaptive light is available (and for shelly bulb RGBW is not)

Attached I added the HomeKit settings.
IMG_0801 = shelly bulb CCT with adaptive light availability on / off
IMG_0799 = shelly bulb RGBW with no adaptive light availability - only color
IMG_0802 = a nanoleaf bulb (on the website they call it RGBCW) with adaptive light availability on / off - and color
IMG_0801
IMG_0799
IMG_0802

InventoryAndi:
Hi. Any ideas / feedback on that?

Issue #186, opened 2020-10-12
Shelly HT


schemhad:
It would be very nice if we could support Shelly HT. Maybe this could be similar to the Power Measuring?

andyblac:
> It would be very nice if we could support Shelly HT. Maybe this could be similar to the Power Measuring?

is HT bluetooth ?

schemhad:
It's an Shelly1 with Humidity and Temperature Sensor. 

https://shelly.cloud/knowledge-base/devices/shelly-ht/

The Sensor could also bought seperate. 

https://shop.shelly.cloud/temperature-sensor-addon-for-shelly-1-1pm-wifi-smart-home-automation#312

andyblac:
ok, wifi, I can't see that battery lasting a week on HomeKit, as it will drain battery as it keep checking connectivity.

schemhad:
But the AddOn Module could run on every Shelly 1 & PM or the retail H&T Module on USB.

andyblac:
> But the AddOn Module could run on every Shelly 1 & PM or the retail H&T Module on USB.

ye I was talking about H&T on battery.

rojer:
yes, this can be done

rojer:
HAP has accessories for temperature and humidity sensors, shouldn't be hard to add.
ideally, we'd auto-detect the sensor on startup so no UI changes would be necessary.

MilanBacaCz:
is there any documentation how to auto detect the temperature and other sensors?

Ludelag:
> yes, this can be done

Dear @rojer does it means will be done it?

rojer:
eventually yes

Ludelag:
> eventually yes

that will be nice

rojer:
looked into this. addon provides a 1-wire bus with output on gpio0 and input on gpio3 (uart0 rx).
output is connected to input (via isolator).
so this can be used for 1-wire sensors or simply as an extra input.

Pixel-Chris:
Will this be usable for the second sensor for the opened position of the garage door opener with a Shelly 1?  (Then I don’t need to power a Shelly 2.5 with an extra power supply)

rojer:
@Pixel-Chris this is indeed possible. the add-on contains circuitry that provides proper isolation for the input.

Pixel-Chris:
Ok 
I must order that addon 😉

Ludelag:
I will for sure also order one and be ready for tasting

[...]

PR #1110, opened 2022-10-30
DHT Sensors

markirb:
This is a first try to address https://github.com/mongoose-os-apps/shelly-homekit/issues/186
It works fine with my DHT.

Things not working so far:
- [x] Discovery of the DHT on first startup (somehow a switch to GDO mode and back on first boot is necessary to show the sensors)
- [x] Webfrontend adaptions


markirb:
Looks better now. Just trying a second time fixed discovery. I do not know why that is...

markirb:
Closing as this is in the other branch now

PR #1145, opened 2022-12-05
Support Dark Mode & UI issues fix

Idomo:
As a follow up to #1083:
- Adding support of Dark Mode.
- Fixing some UI issues.


timoschilling:
Nice! Can you post some screenshots?

Idomo:
Sure:
image

*Also it's obvious to say that this has been tested on mobile-size screens too, but let me know if i missed something.

timoschilling:
nice work, tanks!

PR #1153, opened 2022-12-14
support sensors in gdo mode

### LINKED ISSUE #1102: Shelly 1 Garage opener with DS1820
Hey, i use a few shelly 1 in my House and one in my Garage as door opener.

is it possible with the external plug-in an an DS1820 Sensor to show the Temperature in Homekit?

i can‘t find an solution here

timoschilling:
@markirb what do you think, this should close #1102?


markirb:
not quite, the CreateHAPTemperature Sensor has to be called as well. Also the code to handle non-primary services

markirb:
this is already done in add-on refactoring pr. closing and deleting

PR #1192, opened 2023-01-18
use custom build of mgos/esp32-build docker image

timoschilling:
Many tanks to @markirb for https://github.com/cesanta/mongoose-os/pull/603


markirb:
Nice

PR #1212, opened 2023-02-02
UI polish

timoschilling:


andyblac:
there are lots of formatting issues, double line titles.

PNG image
PNG image
PNG image
PNG image

imo alll the toggle / drop downs need shifting to the right slightly, and make all the info aligned as currently it is not, example below.
PNG image

PR #1235, opened 2023-02-27
implement api toggle for switch and lightbulb

timoschilling:


enji23:
Hello Timo, 
I use the HomeKit and have firmware version 2.11.2 on all my ShellyPlus1. I am currently trying to integrate a Shelly Button 1(https://www.shelly.com/de-ch/products/product-overview/shelly-button-1-white). Since the ShellyPlus1 cannot be controlled directly with the HomeKit firmware, I have used the option via REST-API. 
With "http://xxx.xxx.xxx.xxx/rpc/Shelly.SetState?id=1&type=0&state=%7b%22state%22%3atrue%7d" it works so far.
With "http://xxx.xxx.xxx.xxx/rpc/Shelly.SetState?id=1&type=0&state=%7b%22toggle%22%3atrue%7d" I get an error: {"code":-103, "message": "state is required"}
Am I doing something wrong? 
Many thanks in advance and best regards, Nick

samlitt:
@enji23 I am using this feature for quite some time now and it works great, but since it wasn't in the release back then I compiled from source. But today the long awaited stable release was published, which includes this feature.
Flash version 2.12.1 and try it out, I also tested your url and it works on my fw.

Issue #1024, opened 2022-07-03
Garage door opener initial state


d4rkmen:
Thanks for a graat project first. It's realy something! Shelly devices also are cool

The issue shows afre a power cycle. And actual `closed` gates shows `open`
After very first target state change it sets according to a sensor value.

May be it's related to this line, where the current state characteristic refers to an uninited value `cur_state_`
https://github.com/mongoose-os-apps/shelly-homekit/blob/da78880591d5f2beb62c5d42609796c283b69659/src/shelly_hap_garage_door_opener.cpp#L66

And the actual init happen later here:
https://github.com/mongoose-os-apps/shelly-homekit/blob/da78880591d5f2beb62c5d42609796c283b69659/src/shelly_hap_garage_door_opener.cpp#L98

Thanks in advance

d4rkmen:
It seems the same problem with `tgt_state_` 👀

d4rkmen:
Its sad, but the issue is still present.
I have made some researches. The close sensor of Shelly1 has "normally closed" default config. And by default all works as expected. But set it to "normally open" and restart to reproduce the issue. 
After restart the closed sensor state will be: **open** and only return to real state after actual input event (change), so need to open and close gates before it comes normal.
I have made some loggings:

**First opening**
```
3064698144 shelly_main.cpp:443     Up 3064.69, HAP 0/1/12 ns 1, RAM: 28776/7896; st 0; 5.1: cur:open tgt:open cl:0 op:-1
3066529752 mg_rpc.c:310            Shelly.GetInfoExt via WS_in 192.168.88.125:62260
3067260923 mg_rpc.c:310            Shelly.GetInfoExt via WS_in 192.168.88.125:62260
3068261008 mg_rpc.c:310            Shelly.GetInfoExt via WS_in 192.168.88.125:62260
3068516856 mg_rpc.c:310            Shelly.SetState via WS_in 192.168.88.125:62260
3068523143 shelly_output.cpp:63    Output 1: off -> on (GDO:close)
3068530094 shelly_hap_garage_d:307 GDO 1: Tgt State: open -> closed (0 -> 1) (RPC)
3068536802 shelly_hap_garage_d:222 GDO 1: Cur State: open -> closing (0 -> 3)
3068561880 mg_rpc.c:310            Shelly.GetInfoExt via WS_in 192.168.88.125:62260
3069029458 shelly_output.cpp:63    Output 1: on -> off (pulse_off)
[...]

PR #1311, opened 2023-09-01
Pay attention to GDO sensor mode on init

d4rkmen:
Resolving #1024


d4rkmen:
Did the workflow, flash fw and problem not resolved

PR #1307, opened 2023-08-05
Features/plus addon

markirb:
Try to implement https://github.com/mongoose-os-apps/shelly-homekit/issues/1146

PR #967, opened 2022-03-20
Shelly1/PM: Support Addon as Switch (Input)

markirb:
This would enable an additional Switch via the Addon (e.g. magnetic sensor), for Default mode as well as Garage Door Opener as e.g. requested here:

https://github.com/mongoose-os-apps/shelly-homekit/issues/713

I have nothing to test this with so far, but maybe someone with the appropriate equipment would be able to?

This defaults to momentary switch in Default mode - maybe it should be "Contact Sensor"? But it could be used as many things...

PR #1340, opened 2023-12-01
Addon unification / DHT support / addon as switch / plus addon support

markirb:
This unifies all addon sensor related Pull-requests with a Refactoring and also shoud enable https://github.com/mongoose-os-apps/shelly-homekit/issues/1102

it also adds plusI4 as requested in https://github.com/mongoose-os-apps/shelly-homekit/pull/1307

those are all the other commits:
https://github.com/mongoose-os-apps/shelly-homekit/pull/967 addon as switch
https://github.com/mongoose-os-apps/shelly-homekit/pull/1110 dht
https://github.com/mongoose-os-apps/shelly-homekit/pull/1307 plus addon


markirb:
also: plus addon should work, according to
https://github.com/arendst/Tasmota/discussions/18885
we might need to use GPIO1 instead of GPIO0 (current code) but that is everything

markirb:
@timoschilling can you review this please?

hejsiri:
Does Addon and DHT22 sensor work in Shelly 2PM and beta5 ? It does not detect the sensor in me.

markirb:
It should. Has it worked before. Does it work on original FW?

hejsiri:
OK, I checked on the original software. Addon works correctly with the DHT22 sensor. It shows temprature and humidity. After upgrading back to shelly-homekit beta, the addon is not detected. It does not show the sensor in the webgui.

markirb:
Mhh, you're not on latest beta5. I have reports it works there but:

If it still does not work then the standard mongoose DHT driver seems not to support your model. You can post your exact model, but I think right now I cannot do anything there currently.

hejsiri:
@markirb It is Shelly 2PM.

markirb:
I mean the model of you DHT

hejsiri:
AM2302

markirb:
Totally should work, at least I see no reason why not. Please post this problem as issue with all details. Maybe we find someone who has a similar issue and it can be worked out.

Issue #1013, opened 2022-06-15
Support for Shelly Plus 2PM


herzkerl:
I couldn‘t find any comments regarding the Plus 2PM, is there any news regarding this device?

jonasman:
I also need this

Sorbat:
is there anything new here? can you already say when an alpha/beta version will be available or is planned?

Thank you very much

1974here:
Same here.. longing for a plus2 firmware. 🙂

MagOO33:
Hello, Shelly 2.5 is replaced by Shelly 2PM by resellers in my country because 2.5 has too many hardware problems. It would be very nice to have a mongoose support for this new shelly.

firstusing:
Same here, could only get "plus 2pm´s" for my covers. 
I would too be happy to have a Firmware for it.

c4ris:
a firmware would be nice

AG-dingoant:
I just installed a Shelly 2 Plus. I would really like to have the firmware as well :-)

rojer:
working on it, stay tuned

aqzsed123456:
Hello Rojer, what shelly plus 2 PM will solve problem #985? Thank you very much.

rojer:
who knows, maybe if it's to do with power measurement somehow. though Plus2PM uses the same power measurement chip, so it's unlikely.

aqzsed123456:
Hello rojer, when I do not flash the shelly 2.5 my two large roller shutters work very well with the shelly application on the other hand when I flash them "HomeKit" the calibration hurts so the roller shutters do not close completely. Is it possible to solve this problem by putting a corrective update on shelly 2.5 or to do it on shelly plus 2PM? So that I know if I have to wait for the update of shelly 2.5 or if I need to buy shelly plus 2PM? Thank you very much

solitpapa:
hi, as my 1st from 9 shellys 2.5 after almost 2 years of using starts to die I probably need to start thinking about change to 2PM.
I also would appreciate the introduction of your firmware for that device.

azuma82:
Hello, I have also the new +2PM an will use it for optimized my fans system with the netatmo Wetterstation in Applehome. Thanks for you're work on it

andybenichou:
Hi, thank you for your work it is amazing !! Sorry to ask but do you think the Shelly Plus 2PM will be supported soon, I am using homebridge now but it is not working well? Thank you!

[...]

PR #1341, opened 2023-12-01
Add support of Shelly Plus 2PM

basejumpa:
Contributes to solution of #1013

Actions I did:

- Take development branch [`puls2pm`](https://github.com/mongoose-os-apps/shelly-homekit/tree/plus2pm) @timoschilling created some months ago
- Rebase it to `master`
- Remove `NVS scan` to make it build
- Checked that pin numbers used in the `src/ShellyPlus2PM/shelly_init.cpp` are identical to the ones [spec of device in the tasmota docs](https://templates.blakadder.com/shelly_plus_2PM.html).
- Update over-the-air update of a fresh and new Shelly Plus 2PM with the built I provide at [link removed]
- Integrate it to my Home in two-lights mode.

It works for me.

What still needs to be done from my point of view before merging it to the `master` branch and add it to a release is:

- See if some more of the functionalities of the original software (see [here](https://kb.shelly.cloud/knowledge-base/shelly-plus-2pm)) can/should be ported to homekit
- Adapt documentation (README.md and wiki)

@timoschilling I'd appreciate if you could have a look here. Thx!

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -28,5 +28,13 @@
 	"features": {
 		"docker-in-docker": "20.10",
 		"git": "latest"
-	}
+	},
+	"tasks": [
+		{
+			"label": "Fileserver",
+			"type": "shell",
+			"command": "make build",
+			"problemMatcher": []
+		}
+	]
 }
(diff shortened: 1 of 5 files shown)


timoschilling:
There was a reason that I never published a beta of that!

With that build you can’t revert to stock and can’t use power measurement. So in fake for many usecase of the 2PM you bricked the device!

PR #1419, opened 2024-07-16
Support for Mini 1 Gen 3 / Mini 1PM Gen 3 / Plus 1 Mini

markirb:
TODOS:

- [x] Temp sensor
- [x] Crypto acceleration (hardware MPI)
- [x] In/Output verification
- [x] verify downgrade working  (non PM)

PM support for BL0942 (will be done in separate PR cannot be done without HW available)


metaljay:
How are you getting on with this?

markirb:
It is basically ready. Will be probably released next week as I’m now back from vacation.

PR #1542, opened 2024-10-20
+RGBWPM: support acting as 4 switches & co

markirb:


iyerusad:
Deployed 2.13.2. Tried toggle and edge modes, doesn't appear to do anything. Sticking to Edge mode for now (seems most ideal for testing). Nothing in Debug log when toggle physical switch, tried rebooting Shelly same thing. Software control (web, homekit) successfully works, just physical toggle doesn't appear to doing anything. Confirmed working in Factory firmware last night (before migrating to 2.13.1).

Fairly new to Shelly so apologizes if missing something obvious.

image

Config:
--
- **Shelly +RGBW PM**, 2.13.2 Homekit build
- **Device Profile:** White / "Lights x4" mode. O1 used, O2-O4 HAP hidden
- **24V DC input:** (positive/red into `+` and negative/white into `┴` of Shelly)
- **O1:** (3) LED strips, negatives spliced into one going to `O1` of Shelly
  - Positives of LED strip spliced into `+` port of the Shelly
- **I1:** Line side of Single Pole switch into `I1` of Shelly
  - Load side is spliced into negative/white into `┴` of Shelly

image

iyerusad:
> There is absolutely no change that could cause this... you can downgrade and see if it still works 🤷‍♂️

Let me clarify:
- OEM firmware 1.4.4: Physical switch works/detected actuation
- Homekit 2.13.1: Physical switch doesn't work (what this PR presumably implementing)
- Homekit 2.13.2: Physical switch still doesn't work

In other words, the Homekit version hasn't yet been working. Is there some basics to check for? I can revert to OEM firmware 1.4.4 to test/confirm if needed.

markirb:
Ok now I understand.

This PR does not touch anything regarding the actual inputs. It just introduces a new mode.

This device is a mystery to me. some people say it works some don't...

markirb:
I mean here https://github.com/mongoose-os-apps/shelly-homekit/issues/1556 he had also problems with switching but those were resolved.
Are you sure edge mode is correct? What kind of switch do you have...?

markirb:
Could be that you're right.

I will check in the next day for myself, so we can solve the mystery. (If it was your donation thank you 👍  )

No the change does not affect this, it was only special handling for 100% on, which is now dealt with by the PWM because it has other strange side effects.

markirb:
I tried now with a brand new Shelly RGBW PM everything works as I would expect...

iyerusad:
> I tried now with a brand new Shelly PLUS RGBWPM everything works as I would expect...

Can you share a picture of your wiring? I'm assuming homekit build 2.3.2, white mode, input `I1` is working?

iyerusad:
I just tested following:
1. Revert using homekit web interface from 2.13.2 to ShellyOEM FW 1.4.2
2. Opened the ShellyOEM FW webpage after flash, and Settings -> Device Profile to `Lights x4`.
3. Opened input 1 settings and switched it to "Switch"+"Toggle".
- Its worth noting: The input 1 was showing mode as `Button` + `Single input dimming control - Push to toggle, hold to dim up/down`. It could be possible the mode settings, when set in HomeKit FW, are incorrect -- e.g. is `Edge` in Homekit actually savings as `Button` mode?
5. Flipped physical switch and confirmed working. Tested web toggle and also working.
6. Flashed back to HomeKit FW 2.13.2.
7. Opened HomeKit FW webpage after flash, and set device mode to `White`. Input mode on `I1` is already on `Toggle`.
8. Flipped physical switch and confirmed nothing. Web button activates light. Nothing in the debug logs when toggle physical switch.
---
9. Reverted using homekit web interface from 2.13.2 to ShellyOEM FW 1.4.2
10. Opened the ShellyOEM FW webpage after flash, and Settings -> Device Profile to `Lights x4`.
11. Opened input 1 settings and switched it to "Switch"+"Toggle" (confirmed it was back to `Button`+`Single input dimming control`
12. Flipped physical switch and confirmed working. Tested web toggle and also working.

Seemed to rule out wire issue on my end (I was worried potentially wire splicers might of had some issue, but going back and forth between firmware show the physical switch input as always detected in the OEM fw).

My guess is possible switch configs are incorrectly saved/read when in HomeKit FW, but just a guess.

iyerusad:
> Forget everything: I have a suspicion: Make you Shelly Powerless one time (rebooting is not enought). Afterwards I think it should work.

It started working after completely powering off the Shelly 🎉🎉. Odd. Something that can be patched or is hard requirement of flashing the the firmware?

Only quirk I saw is the toggle switch seems inverted. I'm pretty sure its wired correctly (to where flipping the physical switch on would complete the circuit). Easily enough to workaround, just flip on inverted input, but just heads up in case logic error.

markirb:
The input button problem is basically this:

https://github.com/mongoose-os-apps/shelly-homekit/commit/21a62af327e6fb354e801b97ae81e88044a768ac

iyerusad:
> The input button problem is basically this:
> 
> [21a62af](https://github.com/mongoose-os-apps/shelly-homekit/commit/21a62af327e6fb354e801b97ae81e88044a768ac)

Hmm, don't understand yet enough about the code to grasp what that commit does/doesn't do. 

In any case, sounds like might be worth a note into the flash process to:  flash, wait till it comes online, then completely power off the Shelly (not reboot). [until fixed].
- This will likely more noticed with something like +RGBW because LED power might be not turned off and on during install the same way full high voltage might.

